PR #118, opened 2015-05-16
Translate fixes

bittertini:
Corrected spelling mistakes


NiceandGently:
We use transiflex for translations, we never do change tranlations directly here.
https://www.transifex.com/
Can create a account there and join into mate translations team to fix it, please?
Changes from transifex will merge here than.

bittertini:
Okay :) I have send request to join translation team.

PR #123, opened 2015-06-17
fix html syntax in help translations

NiceandGently:
Those syntax errors prevend that the msgids are merged during complile


monsta:
Surprise... O_o

PR #129, opened 2015-07-08
UI GTK3: use a size request for time plugin dialog window

NiceandGently:
this fixes the not selectable scrolled box in date plugin dialog window

### DIFF
--- a/help/C/index.docbook
+++ b/help/C/index.docbook
@@ -1,15 +1,15 @@
 <?xml version="1.0"?>
 <!DOCTYPE article PUBLIC "-//OASIS//DTD DocBook XML V4.1.2//EN"
 "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd" [
-  <!ENTITY appversion "2.16">
-  <!ENTITY manrevision "2.9">
-  <!ENTITY date "July 2006">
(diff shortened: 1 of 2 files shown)


NiceandGently:
@monsta 
Can you please test ?
Looks like the issue ocours only with gtk+-3.16.

monsta:
Unfortunately, my only GTK+3 VM has 3.16 too...

NiceandGently:
This is OK, because it fixes the issue for you ;)
@infirit told me yesterday that he don't see the issue with gtk+-3.14.
I quess the reason for this is change with srollbars in scrolled windows.

monsta:
Ah yes, they made these stupid appearing-on-mouse-hover scrollbars. Another shit from gnome devs.
I found out that in Synaptic I can't click on the last list item because once I move the mouse there, a horizontal scrollbar appears over the item... :angry:

NiceandGently:
> I found out that in Synaptic I can't click on the last list item because once I move the mouse there, a horizontal scrollbar appears over the item... :angry:

yes, i saw this too with some apps.
OK, in fedora 21 gtk+-3.14 the issue isn't there,
but using this size request doesn't break the window.

Issue #100, opened 2014-11-20
CRLF across 8k boundary inserts an empty line


monnerat:
Originally reported for Fedora 20 at https://bugzilla.redhat.com/show_bug.cgi?id=1166178
Use the attachment at https://bugzilla.redhat.com/attachment.cgi?id=959378 for tests.
Thanks for investigating.
Best regards.
<quote>
Description of problem:
When editing a file with CRLF line ends (DOS), a CRLF across a 8K boundary causes an empty line to be inserted.

Version-Release number of selected component (if applicable):
pluma-1.8.1-2.fc20.x86_64

How reproducible:
Always

Steps to Reproduce:
1. Get the crlfs.txt file in attachment (it does not contain empty line).
2. Edit it with pluma.
3. Browse to the last line: remark it is preceded by a empty line that was not in the initial file.

Actual results:
As mentioned.

Expected results:
No insertion of empty line.
</quote>

sleeveroller:
I can reproduce this issue on Debian Sid - Pluma 1.8.1.

I did some investigation and it seems Pluma reads files in 8k chunks and adds each chunk individually to the gtk text control which tries to tidy up the data as it's added.

When the 2 byte end of line is split on the 8k boundary the gtk control sees the CR at the end of one chunk and the LF at the start of the next chunk as 2 newline indicators.

I don't know enough about the gnome or mate projects to say whether this is a bug in Gtk or in Pluma.

This [patch](https://gist.github.com/sleeveroller/5488bb2414cfd078be80#file-pluma-issue-100-patch) corrects this issue. It's a little hack basically so I haven't created a pull request. But it might serve as a pointer for someone to do a proper fix.

monnerat:
@sleeveroller: thanks for your hint and your work on it. This helped me to produce a cleaner fix. The major difference with your hack is: the carriage return is not stripped, but its processing is delayed as a prefix to the next buffer.

raveit65:
closed via https://github.com/mate-desktop/pluma/pull/148

PR #132, opened 2015-07-10
Fix issue#100: CRLF across 8k boundary inserts an empty line.

monnerat:
The fix works by delaying the handling of a non-initial, trailing carriage-return as a prefix to the next buffer.
See https://github.com/mate-desktop/pluma/issues/100.

### DIFF
--- a/pluma/pluma-document-output-stream.c
+++ b/pluma/pluma-document-output-stream.c
@@ -290,7 +290,6 @@ pluma_document_output_stream_write (GOutputStream            *stream,
 	gsize len;
 	gboolean freetext = FALSE;
 	const gchar *end;
-	gsize nvalid;
 	gboolean valid;
 
 	if (g_cancellable_set_error_if_cancelled (cancellable, error))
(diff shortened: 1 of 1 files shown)


infirit:
merged

infirit:
Reverted as it breaks the tests. It look like line endings are stripped for no reason with this.

```
========================================
   pluma 1.10.2: tests/test-suite.log
========================================

# TOTAL: 5
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  2
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: document-output-stream
============================

/document-output-stream/empty: OK
/document-output-stream/consecutive: OK
/document-output-stream/consecutive_tnewline: **
ERROR:document-output-stream.c:62:test_consecutive_write: assertion failed (inbuf == b): ("hello\rhow\rare\ryou\r" == "hello\rhow\rare\ryou")

FAIL: document-loader
=====================

/document-loader/end-line-stripping: OK
/document-loader/end-new-line-detection: **
ERROR:document-loader.c:76:on_document_loaded: assertion failed (error == NULL): Incomplete UTF-8 sequence in input (g-io-error-quark, 35)
```

monnerat:
Thanks for your tests.
>  It look like line endings are stripped for no reason with this.
I've checked here and noticed I forgot to flush deferred trailing cr at end of data.
However I tried to do it (successfully) at close time, but the tests still fail because the result assertions are made before closing the stream.
Since we need some lookahead to solve the initial problem, we need to require either closing the stream or implement a flush method to be called before the result can be used. I do not see any other solution for it.
Please tell me in which direction you prefer me to investigate.
BTW: the original problem can be exhibited by updating tests/document-output-stream.c (test_consecutive and test_consecutive_tnewline) to use a character count of 3 instead of 2 (all CRs are at the same position mod 2 --> in all cases the error is currently missed).

monnerat:
Closing this one: will submit a new try.

PR #141, opened 2015-08-25
drop support for win32/osx and checkupdate plugin

monsta:
that plugin has never been built in linux anyway due to logic in configure.ac


infirit:
merged

Issue #96, opened 2014-10-30
Pluma is spelled pluma under some circumstances


FuzzyToothpaste:
Occasionally, Pluma will be spelled pluma in the GUI. The only instance of this that I've noticed is in the Help - About dialog where it says "pluma is a small and lightweight text editor for the MATE Desktop".

FuzzyToothpaste:
This makes me wonder what the proper spelling of Pluma even is. I see Pluma and pluma used interchangeably in the program, source code, and even the website and documentation.

infirit:
The correct name is Pluma with a capital P.

monsta:
Fixed that in About dialog in https://github.com/mate-desktop/pluma/commit/8c43c2cfdc87ecbc0058f319012e3c61e31d1629. Any other places without capitalization left? If not, we can close this issue.

FuzzyToothpaste:
You should fix it in the title of the window if that hasn't been done already (I can't test trunk myself), but other than that we should be all good.

monsta:
- fixed titles of main window and prefs window in https://github.com/mate-desktop/pluma/commit/4299e647a7e13118b129266e914fec2a833c44b7
- fixed title of Run command plugin dialog in https://github.com/mate-desktop/pluma/commit/ab4da31e0815c8c7931c7075d7b7c54c6666106e

Hope that's all :smile:

PR #144, opened 2015-09-02
Capitalize pluma desktop entry to match other MATE components

alexandervdm:
I just realized that of all the MATE components present in the menu, pluma and 'dconf-editor' are the [only entries](http://i.imgur.com/u1Dd5k6.png) where the first letter is not capitalized. This commit fixes that for Pluma.


monsta:
See also https://github.com/mate-desktop/pluma/issues/96

infirit:
merged

Issue #90, opened 2014-08-22
please add option to enable "Autocheck Spelling" by default


IndichRMV:
It's frustrating for me to have to enable this option on a case-by-case basis -- I would prefer for spelling errors to be highlighted by default.

The current behaviour is that this setting is remembered per-document, which doesn't make any sense to me -- this is what the Gnome developers implemented when users requested this change; of course, I'm not aware of anyone who was happy with that. See https://bugzilla.gnome.org/show_bug.cgi?id=305055

monsta:
https://github.com/mate-desktop/pluma/issues/146 is merged, so this request is filled.

PR #146, opened 2015-09-23
Add a new Configure dialog for the Spell Check plugin.

sleeveroller:
mate-desktop/pluma#90

(Creating a new pull request here as the previous one was closed when I deleted and recreated cloned branch after some help from @infirit to tidy up my commit)

There are three options for autocheck when a doc is loaded:

Always autocheck, Never autocheck and Remember the autocheck setting for each document

The chosen option is stored in the gsettings schema.
The default setting is to remember by document as this is what Pluma did previously.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -403,6 +403,7 @@ plugins/snippets/Makefile
 plugins/snippets/snippets/Makefile
 plugins/sort/Makefile
 plugins/spell/Makefile
+plugins/spell/org.mate.pluma.plugins.spell.gschema.xml
 plugins/taglist/Makefile
 plugins/time/Makefile
 plugins/time/org.mate.pluma.plugins.time.gschema.xml
--- a/plugins/spell/Makefile.am
+++ b/plugins/spell/Makefile.am
@@ -35,7 +35,7 @@ libspell_la_LDFLAGS = $(PLUGIN_LIBTOOL_FLAGS)
 libspell_la_LIBADD  = $(PLUMA_LIBS) $(ENCHANT_LIBS)
 
 uidir = $(PLUMA_PLUGINS_DATA_DIR)/spell
-ui_DATA = spell-checker.ui languages-dialog.ui
+ui_DATA = spell-checker.ui languages-dialog.ui pluma-spell-setup-dialog.ui
 
 pluma-spell-marshal.h: pluma-spell-marshal.list $(GLIB_GENMARSHAL)
 	$(AM_V_GEN) $(GLIB_GENMARSHAL) $< --header --prefix=pluma_marshal > $@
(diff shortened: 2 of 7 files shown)


monsta:
Seems to work fine, but I have some remarks:

1. The ```Tools -> Autocheck Spelling``` menuitem is still there no matter which option is chosen in this configure dialog. But if "Never autocheck" or "Always autocheck" option is currently selected, that menuitem's setting will be overwritten by it when you open the same document next time - so it's kind of useless in these cases.

2. The ```_``` before "Autocheck spelling on document load" isn't needed :smile:

3. Due to the dialog being non-resizable, the dialog title and the "Autocheck spelling on document load" string are cut.

4. I don't think that "Autocheck" should be capitalized in the option list :smile:


configure-spell-checker

sleeveroller:
@monsta for items 2,3 and 4 I can look into those.

But I think maybe the wording on the new dialog is causing some confusion where item 1 is concerned.
The Tools -> Autocheck Spelling menuitem should always be available regardless of the setting chosen. "Never Autocheck" just means never check on document load. But the user should still be able to toggle the setting on and off within that document session.

Maybe it just needs a more explicit phrase than "Never Autocheck"?

raveit65:
I tested it with gtk3 build, works fine.
But in my opinion the option ' always autocheck' should be enable autocheck immediately without closing/re-open a document or check the menuitem.
This was confusing me :)
Same/vice visa for "never autocheck" when autocheck is enable, here this key should disable the setting.
For 'remember autocheck by document' this behaviour is correct.
Don't know if all this is possible.
> But the user should still be able to toggle the setting on and off within that document session.

sure, this is a MUST

monsta:
> But in my opinion the option ' always autocheck' should be enable autocheck immediately without closing/re-open a document or check the menuitem.

But it's called "Autocheck spelling on document **load**"... :smile:

raveit65:
opps, i missread it

sleeveroller:
Sorry for the slow response here, I was just waiting to see if anyone else made any suggestions and then I forgot about this issue.

Shall I just fix issues 2,3,4 from @monsta above.

If so what is the best way to get those changes in:

Add another commit on top of the original

Redo the commit including the new changes and force push the branch. That will close this pull request and I'll need to create another one.

monsta:
Since these are going to be local fixes for the first commit, it's best to add them, then squash them into the first commit using interactive rebase, and then force-push the new commit.

sleeveroller:
Thanks @monsta 

I've redone this commit now. Let me know if there are still any issues with it.

monsta:
Thanks, merged. We'll handle the remaining issue some time later I think...

PR #180, opened 2016-04-10
plugins: Fix to install externaltools plugin only with Python

obache:
It was accidentally changed to be installed unconditionally with
"drop support for win32/osx".


monsta:
Whoops, my bad :smile:
Merged, thanks!

Issue #188, opened 2016-07-14
[Spell Checker plugin] "Always autocheck" doesn't apply to new documents


iwojia:
Since issue https://github.com/mate-desktop/pluma/issues/90 and PR https://github.com/mate-desktop/pluma/pull/146, the Spell Checker plugin has had a configuration option to "Always autocheck" -- this works when opening any existing document, but does not work when creating a new document....

monsta:
@sleeveroller: can you please take a look at this issue?

monsta:
Fixed in https://github.com/mate-desktop/pluma/commit/e9b5711bc16331ee7de21573f82999053131a9ff.

PR #190, opened 2016-07-20
When 'Always autocheck' option is selected turn on Autocheck for new …

sleeveroller:
mate-desktop/pluma#188

hi @monsta 

I made a small change to try and fix this. Basically the loaded signal is not emitted for new documents so I added a little extra code in the tab added callback function to setup autospell if the document has no uri. I think this signal fires a little too early which is why I had to also add an extra check in set_auto_spell to make sure the doc matches the doc in the view. Without that extra check an error was thrown although functionally it still seemed to work.

Anyway this could definitely do with a review to make sure I haven't broken something else.


monsta:
Looks alright here, thanks. Merged in e9b5711bc16331ee7de21573f82999053131a9ff

Issue #184, opened 2016-06-09
Restore-cursor-position is broken


Bill-MI:
The setting is under **org.mate.pluma -> restore-cursor-position** and defaults in Ubuntu-Mate 16.04 to enabled.  I think most have this same default.  Same buggy behavior in Mate 1.14.1 just verified.

1) Open any text file, I use Caja, hundreds of lines is best to demonstrate the lack of functionality.
2) Place the cursor well down into the file somewhere.
3) Close the file, saved, altered or unsaved - it doesn't matter.
4) Open the file again.
5) Your cursor should be where you left it.  Most the time it is not.

My workaround is use gedit but that's cheating.
Here's what I found that could point to the problem...

If you go to the end of a line and type: (Space)(Backspace)
This makes Pluma think the file has been altered - it hasn't really, of course.
Now, close the Window by clicking "X" or Ctrl+W or Alt+F4.
Answer "Close without saving".
Now it possibly remembers!  Fairly repeatable here.  Generating that dialog is a factor?

I'm not sure how long it's been since this worked right but surprised no one else misses this feature like I do.

raveit65:
I can't really reproduce the issue with mate-1.15.x gtk3 on fedora 24, but maybe this is a random issue, so i will review your PR.

raveit65:
Ok, now i see the different :)
With your PR the scrolled window shows the old cursor position, w/o the cursor is at this position but you need to to type the arrow keys to show the position in the scrolled window.
Nice finder ;)

Bill-MI:
For the record, cursor position is not outside the display window needing a scroll to see it.  Instead, a new cursor position fails to be remembered.

I've adopted a work-around that seems to force cursor position to be remembered for a file:

1) Open file in Pluma, optionally change something.  It doesn't matter, here.
2) Place your cursor where you want it to be remembered for next time.
3) Save the file.  I use Ctrl+S.
4) Type: (space) (backspace).
5) Close Pluma. I use Ctrl+W.  Dialog opens, select "Close without saving".

This seems to work 100%.  Next time this file is opened, the cursor will return exactly where you placed it.  Like it should.

If step 4) is skipped, the cursor position is never remembered and no dialog appears in step 5).  Or, if a cursor position was previously remembered, it doesn't update to the new location as it should.

This is very repeatable in Ubuntu-Mate 16.04 with Mate 1.12 or 1.14.

Does anyone know where this data is stored?  I tried to find it unsuccessfully.

[...]

PR #191, opened 2016-07-21
tab: restore the cursor and scroll in an idle.

sleeveroller:
mate-desktop/pluma#184

I found this commit in the gedit git repository which seems to fix this problem so I've ported it to pluma:

(https://github.com/GNOME/gedit/commit/cb5829d81cd5a48e471ce526e3dd01842a80b703)

From my testing it looks like the cursor was being correctly positioned but depending on the timing it would not always scroll down to the cursor. This change basically waits to make sure the tab is loaded properly before scrolling down to the cursor.

### DIFF
--- a/pluma/pluma-tab.c
+++ b/pluma/pluma-tab.c
@@ -84,6 +84,8 @@ struct _PlumaTabPrivate
 	gint                    auto_save : 1;
 
 	gint                    ask_if_externally_modified : 1;
+	
+	guint			idle_scroll;
 };
 
 #if GTK_CHECK_VERSION (3, 0, 0)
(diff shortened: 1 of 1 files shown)


raveit65:
Looks good for me with gtk3 build.
@monsta @XRevan86 
Can you do a code review please?
My results:
```
Ok, now i see the different :)
With your PR the scrolled window shows the old cursor position, w/o the cursor is at this position but you need to to type the arrow keys to show the position in the scrolled window.
Nice finder ;)
```

monsta:
Well, I couldn't reproduce that race condition, but things seem to be ok with this PR, so I think it's alright to merge it.

raveit65:
OK, for me it is a big improvement.

raveit65:
merged https://github.com/mate-desktop/pluma/commit/01e91d300b73e178130cf41d5356122bb8146212
Thank you

PR #193, opened 2016-07-30
deprecations

raveit65:
please review

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -173,8 +173,6 @@ case "$with_gtk" in
 esac
 AC_SUBST(GTK_API_VERSION)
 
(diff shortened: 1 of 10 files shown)


lukefromdc:
All the UI elements are rendering normally, I don't see any issues that came from this PR and have been using it a couple weeks.

I've had occasional issues most of the summer with Pluma going nonresponsive but I'm pretty sure that predates this PR. That issue is too infrequent to invoke at will for debugging, happens just often enough to be a nuisance when writing software, actlng like Gedit used to when the file browser from which a file had been opened was killed.

monsta:
GTK+2 build is ok (well, it's not really affected by these changes).

monsta:
All good in GTK+3 with 3.20 as well.

raveit65:
merged.

PR #201, opened 2016-11-06
move to GTK+3 and drop Python plugins support

monsta:
@lukefromdc @sc0w @raveit65 
Well, this should work as before in GTK+3 builds, it's more like formal testing before I start working on backporting libpeas support from gedit.

Python plugins are left as is for now - they'll be ported to libpeas later.


lukefromdc:
Seems to work fine on my main system in a quick test. Not using any Python plugins here so that's a non-issue, but won't be for everyone.

PR #205, opened 2016-12-05
port plugin system and C plugins to libpeas

monsta:
Please test it if you can guys.
You'll need to install dev files of libpeas and libpeas-gtk. They actually might be in a single package in your distros - for example, it's so in Debian/Ubuntu, package is named simply libpeas-dev.

Porting Python plugins to libpeas will be done later.

Just in case - how to test plugins:

* changecase: select some text and use ```Edit -> Change Case``` menu item

* docinfo: use ```Tools -> Document Statistics``` menu item

- filebrowser:
  - press F9 to enable side pane
  - select the tab with file manager icon there to reveal the filebrowser window

* modelines:
  - set default tab width to 4 in editor preferences
  - open a source file that has modeline with tab width set to 8 (example:
```plugins/modelines/modeline-parser.c``` - it has such modeline at the bottom)
  - check tab width indicator on the statusbar - tab width should be set to 8 for that file

- sort: use ```Edit -> Sort...``` menu item

* spell:
  - use ```Tools -> Check Spelling...``` menu item
  - use ```Tools -> Autocheck Spelling``` menu item
  - in the plugin list, make sure Preferences button is enabled and works for this plugin
  - test various options in plugin's preferences dialog

- taglist:
  - press F9 to enable side pane
  - select the tab with plus icon there to reveal the taglist window

* time:
  - use ```Edit -> Insert Date and Time...``` menu item
  - in the plugin list, make sure Preferences button is enabled and works for this plugin
  - test various options in plugin's preferences dialog

- trailsave: add some trailing spaces to any file, save the file - trailing spaces should be gone


lukefromdc:
I just tested these, all of the plugins seem to work. Test was with GTK 3.22

PR #208, opened 2016-12-19
port Python plugins (except for snippets) to gi and libpeas

monsta:
Please test it if you can guys.
You'll need introspection packages for libpeas and Pluma itself, and also Python plugins loader from libpeas. Depending on your distro, the loader might be in the same package as libpeas itself or in a separate one.

Just in case - how to test plugins:

* externaltools:
  * press Ctrl-F9 to enable the bottom pane
  * select the tab with the gear icon to reveal the shell output window
  * use ```Tools -> External Tools``` menu (or use keyboard shortcuts) to run various tools
  * use ```Tools -> Manage External Tools...``` menu item to invoke the tools manager dialog
  * try adding and removing tools (they should be saved to ```~/.config/pluma/tools/```)
  * try changing various settings for the existing tools (code, shortcuts, languages and so on)

- pythonconsole:
  - press Ctrl-F9 to enable the bottom pane
  - select the tab with Python icon to reveal the Python console

* quickopen:
  * press Ctrl-Alt-O to invoke the quick open dialog
  * in the search field, type folder/file names that are in your home folder or in ```~/.gtk-bookmarks```
  * use arrows and Enter to navigate the folders and files that appear in the list

As for the snippets plugin, I'm still trying to make it work (code completion doesn't work at all at the moment), so it will be kept in a separate dev branch for a while.

Known issue: the bottom pane of Pluma doesn't look good in GTK+3 build. Nobody could find it earlier because there were no plugins that used it, so it was simply disabled in the menu... I will try to fix it later.

Also if you have a GTK+2 build of Pluma somewhere, it would be nice to check if there are any regressions in these plugins, compared to GTK+2 build. I've already tested that in a VM but I could miss something.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -187,6 +187,8 @@ else
 	have_introspection=no
 fi
 
+AM_PATH_PYTHON([2.7])
+
 dnl ================================================================
 dnl GSettings related settings
 dnl ================================================================
@@ -252,6 +254,10 @@ pixmaps/Makefile
 plugins/Makefile
 plugins/changecase/Makefile
 plugins/docinfo/Makefile
+plugins/externaltools/data/Makefile
+plugins/externaltools/Makefile
+plugins/externaltools/scripts/Makefile
+plugins/externaltools/tools/Makefile
 plugins/filebrowser/Makefile
 plugins/filebrowser/org.mate.pluma.plugins.filebrowser.gschema.xml
 plugins/modelines/Makefile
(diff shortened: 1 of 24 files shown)


lukefromdc:
External tools works, tools such as open in terminal seem to work, and "build" on a random file gives No Makefile found! as output, implying these are working. 

Not sure what to do with the python  console but it opens, you get the two tabs between  "Shell Output" and "Python Console."

Quickopen works, these seem to be good to go as far as I can tell

PR #214, opened 2017-02-24
port snippets plugin to gi and libpeas

monsta:
Use ```Tools -> Manage Snippets...``` menu item to open Snippets Manager. You can check the available snippets for various languages there.

How to test it: open some file with code, type one of supported snippets and press Tab.
Example: type ```for``` in C file and you should get this:

snippets-1

Then you can change the values in placeholders to something else. Use Tab and Shift-Tab to move between placeholders.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -250,6 +250,10 @@ plugins/pythonconsole/Makefile
 plugins/pythonconsole/pythonconsole/Makefile
 plugins/quickopen/Makefile
 plugins/quickopen/quickopen/Makefile
+plugins/snippets/data/lang/Makefile
+plugins/snippets/data/Makefile
+plugins/snippets/Makefile
+plugins/snippets/snippets/Makefile
 plugins/sort/Makefile
 plugins/spell/Makefile
 plugins/spell/org.mate.pluma.plugins.spell.gschema.xml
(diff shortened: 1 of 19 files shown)


lukefromdc:
Without this PR snippets doesn't even show up-we need this to re-enable snippets

raveit65:
hmm, snipset plugin was disabled for 1.17.3 tarball.
......a hint for creating a new tarball wouldn't be helpful.

monsta:
Added a fix for an old visual issue which is present in 1.16 (GTK+2 build) as well, so it will go into backports later. The issue is easily reproducible:

- type ```for```, press Tab

snippets-2

- press Ctrl-Z

snippets-3

After that, you'll have to use Ctrl-Z a few times more to get rid of these remains, and Tab completion might stop working completely until you clean all that and type ```for``` on a new line :slightly_smiling_face:

raveit65:
@monsta 
Can you please rebase against master to have the version bump in this branch?

monsta:
Rebased, should be with 1.17.4 now

raveit65:
Works well, with the full PR is see only one remaining **for** after pressing crtl+z.
I guess this is the expected behaviour?

monsta:
Yes, and then you can press Tab again to get the snippet appear again - previously it didn't always work properly after undo.

PR #225, opened 2017-04-25
Support new location of gtk bookmarks file

raveit65:
The gtk bookmarks file is now in XDG dir, so let's first try there and
otherwise fall back to the old location.

Please test, commit didn't apply clean.

### DIFF
--- a/plugins/filebrowser/pluma-file-bookmarks-store.c
+++ b/plugins/filebrowser/pluma-file-bookmarks-store.c
@@ -494,66 +494,103 @@ add_bookmark (PlumaFileBookmarksStore * model,
 	return ret;
 }
 
-static void
-init_bookmarks (PlumaFileBookmarksStore * model)
+static gchar *
+get_bookmarks_file (void)
 {
-	gchar *bookmarks;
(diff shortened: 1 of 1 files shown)


raveit65:
Indeed, adding or removing one bookmark from ~.config/gtk-3.0/bookmarks remove all bookmarks from filebrowser-plugin in sidebar.
But filebrowser-plugin don't crash here and is usable.
And using the reload button brings back all existing bookmarks, without closing pluma or enable/disable the filebrowser plugin :-)
I see same behaviour with using old ~gtk-bookmarks file **before** this PR.
You can test this with simply removing or adding a bookmark line from this file with pluma.
So, i see no regression here.

lukefromdc:
https://github.com/raveit65 The whole plugin does NOT crash, what crashed was the list of bookmarks only, same as your results. If this is same as before this PR looks ready to go and we have a new (well actually old) issue to deal with

raveit65:
There is a toolbar on top of the sidebar with a reload button, which reloads the bookmarks here.
Does that work on your system?

lukefromdc:
Reloading the bookmarks works fine

PR #227, opened 2017-05-13
deprecations and warnings

raveit65:
Please test.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -157,7 +157,7 @@ PKG_CHECK_MODULES(PLUMA, [
 	glib-2.0 >= 2.36.0
 	$GMODULE_ADD
 	gthread-2.0 >= 2.13.0
-	gio-2.0 >= 2.26.0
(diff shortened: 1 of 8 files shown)


raveit65:
pluma-notebook: don't use deprecated gdk_cursor_new is updated after a rebase against master.

lukefromdc:
Rebased version also works fine

monsta:
Since gio is a part of glib, https://github.com/mate-desktop/pluma/commit/91834ee958566d4edc2af9ecb90ac4440765c7d1 essentially bumps glib requirement to 2.37 instead of 2.36.
But... GTK+ 3.14 already requires 2.41.2 - so we're actually safe to bump the requirement to 2.42 (and drop some old code). :slightly_smiling_face:

raveit65:
Thanks for info, i can update requirements next time but to decide which code is old and can be dropped is definitely a job for you ;-)

PR #235, opened 2017-06-23
disable the abbility to save the document if it has no changes

### LINKED ISSUE #199: unsaved file indication
WIBNI
the filename being edited were highlighted if unsaved (a la geany) and the save button were greyed out once the current file was saved.

sc0w:
this prevent us the write to disk uselessly

Closes #199 

we can see this same behavior in geany


raveit65:
Thank you

PR #237, opened 2017-07-09
avoid deprecated gdk_display_get_screen and gdk_screen_get_number

sc0w:


raveit65:
I think we need GTK_VERSION_CHECK.
gdk_screen_get_number has been deprecated since version 3.22
https://developer.gnome.org/gdk3/stable/GdkScreen.html#gdk-screen-get-number
gdk_display_get_screen has been deprecated since version 3.20
https://developer.gnome.org/gdk3/stable/GdkDisplay.html#gdk-display-get-screen
Sadly, they use different versions.
@monsta @XRevan86 
thoughts?

sc0w:
@raveit65 done!

monsta:
[Docs](https://developer.gnome.org/gdk3/stable/GdkDisplay.html#gdk-display-get-default-screen) say `gdk_display_get_default_screen` can be used since 2.2.

monsta:
Also this changes the format of the internal messages that some plugins might use; did you check that all the plugins still work fine?

sc0w:
> Docs say gdk_display_get_default_screen can be used since 2.2.

yes, I think this must work without the checks

Commit rebased again

> Also this changes the format of the internal messages that some plugins might use; did you check that all the plugins still work fine?

plugins:

"external tools", "python console", "quick open" and "snippets" don't work with the tooltip message: "plugin loader 'python' was not found", so, I think it isn't related with the PR

"modelines", "save without trailing spaces" and "tag list", I am not sure how to test, can someone test it?

the other plugins are working fine here

monsta:
Ah dammit, lack of dependencies... you'll need to install `libpeas-1.0-python2loader` package 🙂

monsta:
I think I'll check the code to find out which plugins actually use the message bus and where/when they do it...

sc0w:
commit renamed and rebased with all the other deprecations (display_get_screen/gdk_screen_get_number) in pluma

and, I am not sure, we need the checks for gtk in pluma.c?

sc0w:
@monsta 

PR updated, and now, I think doesn't change the format of the internal messages that some plugins might use

raveit65:
Thank you

PR #239, opened 2017-07-28
Dev gtk show uri

raveit65:
Please test

### DIFF
--- a/plugins/filebrowser/pluma-file-browser-widget.c
+++ b/plugins/filebrowser/pluma-file-browser-widget.c
@@ -2473,7 +2473,11 @@ directory_open (PlumaFileBrowserWidget *obj,
 	if (FILE_IS_DIR (flags)) {
 		result = TRUE;
 
+#if GTK_CHECK_VERSION (3, 22, 0)
+		if (!gtk_show_uri_on_window (NULL, uri, GDK_CURRENT_TIME, &error)) {
+#else
 		if (!gtk_show_uri (gtk_widget_get_screen (GTK_WIDGET (obj)), uri, GDK_CURRENT_TIME, &error)) {
+#endif
 			g_signal_emit (obj, signals[ERROR], 0,
 				       PLUMA_FILE_BROWSER_ERROR_OPEN_DIRECTORY,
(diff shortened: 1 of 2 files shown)


lukefromdc:
Here's what the redraw issue looks like, and I just verified this does not occur in Master:
pluma_sidebar_redraw_mess 2

And on changing themes after opening a folder in Caja from the Pluma sidebar:

pluma_sidebar_redraw_mess

raveit65:
Ok, confirmed with ubuntu-legacy and Menta themes.
Much worse is the flickering of the treeview of the the sidebar if you hover with the mouse.
I think i will split this PR.
.....sometimes i really hate gtk+ changes,.....well it's only software :-)

lukefromdc:
Help works fine after latest change, concur with using that as a separate PR. The other part still makes a mess of the sidebar.

raveit65:
Using NULL for file-browser-widget seems to make the different. No redraw issue here any more.
Please test again.

lukefromdc:
This works, and fixes the redraw bug. 

Strange that GTK would work this way, I would not have expected NULL to work at all, but had no alternative in my original tests on the panel because the functions involved had no access to any of the possible windows. That left NULL as the only solution and it not only worked but fixed these bugs. In the GTK documentation for the older deprecated gtk_show_uri we have this paragraph:

"A convenience function for launching the default application to show the uri. Like gtk_show_uri_on_window(), but takes a screen as transient parent instead of a window."

Perhaps that "transient parent" is not ever used to actually show the window called and is only used for some kind of memory/structure management internally, otherwise being mostly a source of bugs? I'm just glad NULL works instead of segfaulting over this.

monsta:
By any change, did you accidentally pass some GtkWidget that is not a GtkWindow as the first argument?
Looking at the old code, I think it probably was PlumaFileBrowserWidget which is inherited from GtkBox, so it's not a GtkWindow.

monsta:
Using NULL is explicitly allowed in the docs, as I mentioned in https://github.com/mate-desktop/mate-power-manager/pull/223 (see the `[allow-none]` note).

Issue #211, opened 2017-01-19
[Feature Request]: Switch tabs with CTRL+TAB [$30]


epgfm:
It'd be nice to be able to switch open tabs the same way we can for Firefox: CTRL+TAB. and CTRL+SHIFT+TAB for reverse switch.
Currently we can only switch using ALT+SHIFT+NUMBER (so it's absolute switching with a hard limit of 10 tabs)

samtygier:
Pluma can switch tabs with CTRL+ALT+PgUp and CTRL+ALT+PgDn
Mate Terminal and Caja use CTRL+PgUp and CTRL+PgDn

So should should they all be changed to CTRL+TAB and CTRL+SHIFT+TAB to match Firefox? or should CTRL+TAB and CTRL+SHIFT+TAB be added as additional keyboard shortcuts? If so should pluma drop the need for ALT?

epgfm:
The way I see it, the Firefox shortcuts are now the expected, "natural" ones.
If someone really wants to hang on to the current shortcuts, maybe a way to configure the keybinding would be better.

samtygier:
Looking into this it seems like gtk reserves CTRL+TAB for changing focus between widgets. https://bugzilla.gnome.org/show_bug.cgi?id=123994 . So for example if you have the text area focused, then CTRL+TAB lets you focus the tool bar. There may be uses who can't use a mouse dependant on this.

As mentioned on that bug, pidgin bypasses this, by intercepting all keypress and handling the CTRL+TAB case. So it is possible, just a little hacky.

raveit65:
CTRL+TAB is used for switching between applications too. So this can be confuse if pluma isn't a maximize window and isn't focus.
In result you need to use CTRL+TAB 2 times. One time for selecting the plum window and than for switching the tabs.
So i would prefer to use the same keys like caja or mate-terminal.
Btw. not everyone is using firefox, so why follow them?

epgfm:
You're using CTRL+TAB to switch applications? I thought the universal shortcut for that was ALT+TAB. Anyway, that's not only Firefox using CTRL+TAB to switch between panel, Google Chrome does too, and Opera.

raveit65:
I edid previous comment.

raveit65:
Btw. volunteers welcome

raveit65:
I agree with follow using firefox shortcuts, but using both the old ones and new ones seems a bit overload for me.
For all 3 applications.

epgfm:
[...]

PR #242, opened 2017-08-05
Use CTRL+TAB and CTRL+SHIFT+TAB for tab switching 

samtygier:
Solution for #211

Use the more widely used keyboard shortcuts for tab switching.

Note that GTK uses CTRL+TAB and CTRL+SHIFT+TAB for changing widget
focus, so the keypress signals must be caught directly rather than using
GtkAction.

### DIFF
--- a/pluma/pluma-ui.h
+++ b/pluma/pluma-ui.h
@@ -128,9 +128,12 @@ static const GtkActionEntry pluma_menu_entries[] =
 	  N_("Save all open files"), G_CALLBACK (_pluma_cmd_file_save_all) },
 	{ "FileCloseAll", GTK_STOCK_CLOSE, N_("_Close All"), "<shift><control>W",
 	  N_("Close all open files"), G_CALLBACK (_pluma_cmd_file_close_all) },
-	{ "DocumentsPreviousDocument", NULL, N_("_Previous Document"), "<alt><control>Page_Up",
+	/* Note Next and Previous actually handled in handle_keypress() due to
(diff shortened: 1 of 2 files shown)


samtygier:
This causes a warning:

```
pluma-window.c: In function ‘handle_keypress’:
pluma-window.c:3867:44: warning: passing argument 1 of ‘_pluma_cmd_documents_previous_document’ from incompatible pointer type [-Wincompatible-pointer-types]
     _pluma_cmd_documents_previous_document(event, window);
                                            ^~~~~
```
`_pluma_cmd_documents_previous_document()` can also be called from the menu item via a GTKAction and so expects `event` to be a `GtkAction`. From `handle_keypress()` we have a `GdkEventKey`. Not sure what's best to do, given that `_pluma_cmd_documents_previous_document()` doesn't do anything with `event`.

sc0w:
- I see two build warnings:

```
pluma-window.c:3867:44: warning: passing argument 1 of ‘_pluma_cmd_documents_previous_document’ from incompatible pointer type [-Wincompatible-pointer-types]
     _pluma_cmd_documents_previous_document(event, window);
```

```
pluma-window.c:3869:40: warning: passing argument 1 of ‘_pluma_cmd_documents_next_document’ from incompatible pointer type [-Wincompatible-pointer-types]
     _pluma_cmd_documents_next_document(event, window);
```

- The PR seems to work, and the user can change between tabs with ctrl+tab/ctrl+shift+tab or ctrl+alt+avpg/ctrl+alt+repg

- but, how to access to the toolbar only with keyboard? That feature was removed with this PR

Issue #250, opened 2017-09-25
wrong behavior holding down [middle + right] click


sc0w:
how to reproduce:

in the red area:

![image4493](https://cloud.githubusercontent.com/assets/7734191/26037443/ce61d162-38f2-11e7-9616-8cc59ac650ce.png)

1. holding pressed middle click

2. right click while the middle click is pressed

#### Actual behaviour

![rightclickwrong](https://user-images.githubusercontent.com/7734191/30797416-3e4f2898-a1d6-11e7-81d1-37517a3c94c6.png)

#### Expected behaviour

Nothing to do

(pluma 1.18.2 and latest git master)

sc0w:
see the video:

[out-47.ogv.zip](https://github.com/mate-desktop/pluma/files/1328919/out-47.ogv.zip)

PR #229, opened 2017-05-15
Fix: pluma closes tabs unexpectedly with middle click

sc0w:
Fixes #228

PR #251, opened 2017-09-25
Fix: wrong behavior holding down [middle + right] click

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/250

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -56,6 +56,8 @@
 
 #define PLUMA_VIEW_GET_PRIVATE(object)(G_TYPE_INSTANCE_GET_PRIVATE ((object), PLUMA_TYPE_VIEW, PlumaViewPrivate))
 
+static gboolean middledown = FALSE;
+
 typedef enum
 {
 	GOTO_LINE,
@@ -121,7 +123,10 @@ static gboolean pluma_view_drag_drop		(GtkWidget        *widget,
 	      					 gint              x,
 	      					 gint              y,
 	      					 guint             timestamp);
+
 static gboolean	pluma_view_button_press_event	(GtkWidget        *widget,
+						 GdkEventButton   *event);
+static gboolean	pluma_view_button_release_event	(GtkWidget        *widget,
 						 GdkEventButton   *event);
 
 static gboolean start_interactive_search	(PlumaView        *view);
@@ -206,6 +211,7 @@ pluma_view_class_init (PlumaViewClass *klass)
 	widget_class->drag_data_received = pluma_view_drag_data_received;
 	widget_class->drag_drop = pluma_view_drag_drop;
 	widget_class->button_press_event = pluma_view_button_press_event;
+	widget_class->button_release_event = pluma_view_button_release_event;
 	klass->start_interactive_search = start_interactive_search;
 	klass->start_interactive_goto_line = start_interactive_goto_line;
 	klass->reset_searched_text = reset_searched_text;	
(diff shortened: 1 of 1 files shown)


pluma/pluma-view.c
```diff
@@ -2007,6 +2023,17 @@ pluma_view_button_press_event (GtkWidget *widget, GdkEventButton *event)
...
 
+static gboolean
+pluma_view_button_release_event (GtkWidget *widget, GdkEventButton *event)
+{
+	if (event->button == 2)
+	{
+		middledown = FALSE;
+	}
+
+	return GTK_WIDGET_CLASS (pluma_view_parent_class)->button_press_event (widget, event);
```
monsta: Maybe it should be `button_release_event` since we're in similarly named function here?

pluma/pluma-view.c
```diff
@@ -56,6 +56,8 @@
 
 #define PLUMA_VIEW_GET_PRIVATE(object)(G_TYPE_INSTANCE_GET_PRIVATE ((object), PLUMA_TYPE_VIEW, PlumaViewPrivate))
 
+static gboolean middledown = FALSE;
```
monsta: I think it would be better to add this variable to the private struct (`PlumaViewPrivate`) and use a bit more descriptive name, like `middle_button_pressed`.

sc0w:
> Maybe it should be button_release_event since we're in similarly named function here?

oops, yes, commit rebased

> I think it would be better to add this variable to the private struct (PlumaViewPrivate) and use a bit more descriptive name, like middle_button_pressed.

I'm fine with this code, in the past I did similar PR: https://github.com/mate-desktop/pluma/pull/229

monsta:
It's a common practice to keep local variables in the private struct. I don't know how many PlumaView's could be there.

sc0w:
@monsta 

I don't know how to do that

I tried, but it doesn't work, did you try?

monsta:
Let's see... the GtkWidget which is passed as the first argument to the event handlers is most probably the PlumaView itself. So something like this might work:

```C
PlumaView *view = PLUMA_VIEW (widget);
view->priv->middle_button_pressed = TRUE;
```

(I didn't test yet)

sc0w:
@raveit65 @lukefromdc please test, review, thoughts, opinions...

lukefromdc:
A test build worked, suppressing the undesired appearance of the tab top's the right-click menu. I haven't taken a close enough look at the rest of the Pluma coding to evaluate the issue of variable naming/keeping in the private struct here. Worked in a quick test here, but I suspect keeping the closest possible control of variables would make future work easier and less likely to run into surprises.

(detailed thoughts on this below, not specific to this one PR and not to block anything)

One thing about a complex project in C with many authors and a long history is that it is really easy for one programmer to generate issues for another, later programmer, or to get into trouble because a previous programmer did something unexpected. We need to remember that we are all working in what many say is the language in which it is easiest to get into trouble, though it also gives far faster finished code than things like GNOME's beloved javascript. In short, watching our coding practices and checking each other is more valuable in C than it is in a lot of other languages.

When things don't work after being done the way that is considered best practices, it may pay to look at why it is not working. Did a previous programmner's kludge in the same place come back to bite a later effort to do things right? If fixing that is too complex or beyond our ability(say because the whole file has problems), then we may need to kludge it ourselves, but should document that with a FIXME.  

On the other hand, we should not have to refactor half of a file to deal with one minor fix either, that would cause development to slow to a crawl-and then probably bring in lots of new bugs, the way the GtkApplication transition did not only for Caja but also for Nautilus. Those were incredibly complex, and GNOME's refactoring of Nautilus forced by that change took them years to debug.

raveit65:
> It's a common practice to keep local variables in the private struct. I don't know how many PlumaView's could be there.

I don't know what is better, but please don't keep a working easy PR a long time open only for a coding style question.

lukefromdc:
I agree: merge this and get it out of the way

monsta:
No problem, merge it, I'll work on it later.

raveit65:
@monsta 

> No problem, merge it, I'll work on it later.

Please merge it for youself instead of writing this text and asking other people to do that. ;-)
We are all busy....

raveit65:
merged https://github.com/mate-desktop/pluma/commit/4df44887f4b7547fc9b31d1cb3f61688579fe39b
Thanks

Issue #230, opened 2017-05-17
Do not request to save unsaved, empty, nonexisted files. 


thebombzen:
From the command line, if you type `pluma blahblah` (or rather, open any other non-existent file), pluma will create a blank document but won't create the file. If you then immediately close the nonexistent file you just opened, pluma will ask if you want to save your work.

This behavior might be intended but it's very annoying. This happens a lot of you accidentally make a typo in the name of the file you want to open. If you accidentally open `makefile` instead of `Makefile`, and say, "Oops, that's a typo" and close it, it shouldn't ask if you want to save `makefile`, because you don't want to save it. It's empty, you made no changes, and it never existed.

monsta:
Yeah, gotta fix it. I'm stumbling upon this all the time when I make a typo in some filename...

monsta:
@sc0w: can you please have a look at this?

PR #252, opened 2017-09-25
don't request to save documents with no changes

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/230

@monsta please test


raveit65:
> Works for me. Asked other devs to review just in case.

@monsta 
Maybe you don't know that we're a small team ;-)
2 approvals are really enough for just a small change

Issue #224, opened 2017-04-21
[Feature Request] - Keyboard Shortcut to increase font size. [$20]


epgfm:
It'd be nice to bind [CTRL] + [+] to a 1 point increase in font size, the same shortcut used in Firefox and mate-terminal.
Similarly, bind [CTRL] + [-] to a 1 point decrease.

PR #253, opened 2017-09-28
[ctrl] + [+] / [ctrl] + [-] to increase/decrease font size

sc0w:
Closes https://github.com/mate-desktop/pluma/issues/224

The feature is available if "use the system fixed width font" is disabled in preferences

### DIFF
--- a/pluma/pluma-window.c
+++ b/pluma/pluma-window.c
@@ -309,6 +309,45 @@ pluma_window_key_press_event (GtkWidget   *widget,
 	GtkWindow *window = GTK_WINDOW (widget);
 	gboolean handled = FALSE;
 
+	if (event->state & GDK_CONTROL_MASK)
+	{
+		GSettings *settings;
+		gchar     *font;
+		gchar     *tempsize;
+		gint       nsize;
+
+		settings = g_settings_new ("org.mate.pluma");
+		font = g_settings_get_string (settings, "editor-font");
+		tempsize = g_strdup (font);
+
+		g_strreverse (tempsize);
+		g_strcanon (tempsize, "1234567890", '\0');
+		g_strreverse (tempsize);
+
+		gchar tempfont [strlen (font)];
+		strcpy (tempfont, font);
+		tempfont [strlen (font) - strlen (tempsize)] = 0;
+
+		sscanf (tempsize, "%d", &nsize);
+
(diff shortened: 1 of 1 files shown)


monsta:
I'm not sure about this... these string manipulations don't seem reliable to me. Maybe try [PangoFontDescription functions](https://developer.gnome.org/pango/stable/pango-Fonts.html)? They're used in mate-terminal.

There's also Ctrl+0 shortcut available in both mate-terminal and Firefox (mentioned by @epgfm in #224), it resets zoom to default, which can be useful too.

sc0w:
@monsta 

> I'm not sure about this... these string manipulations don't seem reliable to me. Maybe try PangoFontDescription functions? They're used in mate-terminal.

I am completely sure, it works with all fonts, there is no reason to work with your suggestion for me

> There's also Ctrl+0 shortcut available in both mate-terminal and Firefox (mentioned by @epgfm in #224), it resets zoom to default, which can be useful too.

This PR closes that issue, I can't see the mentioned [ctrl] + [0], maybe in another PR

@raveit65 @lukefromdc please test, review, thoughts, opinions...

lukefromdc:
I'm no expert on string handling in C, but it does look like this was done using recommended practices for safe string handling(using safer g library versions of string functions where available, working on duplicate strings, etc). About to test...

lukefromdc:
In my test this worked with the + (shift =) and - keys on the top of the keyboard but did NOT work with the + and - keys on the numeric keypad. I checked the function of those keys by typing from them into a pluma window and they delivered the correct characters.

sc0w:
@lukefromdc thanks for testing

> In my test this worked with the + (shift =) and - keys on the top of the keyboard but did NOT work with the + and - keys on the numeric keypad.

yes, in mate-terminal it doesn't work with the numeric keypad, I have no idea how to work that

raveit65:
Having a crtl+0 keybinding like in mate-terminal would be nice, but this can be done in another commit.
No reason for not merging this.
For me this works OK, no idea why @monsta wants to have it in complete other way....

raveit65:
merged https://github.com/mate-desktop/pluma/commit/f0b6b6693720aef9281c840b84deff9a1a6e1aa2
Thank you.

> If you're in a hurry, merge it, I'll replace the code later then.

Great, looks like you have tons of free time for MATE in future :-)

Issue #217, opened 2017-03-13
Selected text is lost when selecting text in another program/window


ZrN-kAyDqY:
Pluma 1.16.0 on Arch:
1) Start a Pluma document with text     123456789
2) Highlight the following text     123
3) Open a webpage in Firefox, or document in LibreOffice Writer, and select some text.
4) 123 is no longer selected in Pluma.  
{Go back to Pluma so the window has focus, to verify} 

Is this really the same as #179 and nothing can be done about it?  Selecting text in Firefox first and then going to Pluma to select text does not display the behavior.  Selecting text in Writer first and then going to Pluma to select text does not display the behavior.   If this were working it is helpful when doing various copy/pastes from the Pluma document to some other app, and keeping track of last copied text etc.

PR #255, opened 2017-10-17
Fix: selected text is lost when selecting text in other program

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/217

### DIFF
--- a/pluma/pluma-window.c
+++ b/pluma/pluma-window.c
@@ -60,6 +60,7 @@
 #include "pluma-enum-types.h"
 #include "pluma-dirs.h"
 #include "pluma-status-combo-box.h"
+#include "dialogs/pluma-search-dialog.h"
 
 #define LANGUAGE_NONE (const gchar *)"LangNone"
 #define PLUMA_UIFILE "pluma-ui.xml"
@@ -72,7 +73,9 @@
 					 PlumaWindowPrivate))
 
 /* Local variables */
-static gboolean cansave = TRUE;
+static gboolean       cansave = TRUE;
+static GtkTextIter    selection_end = {0};
+static PlumaDocument *doc_selection = NULL;
 
 /* Signals */
 enum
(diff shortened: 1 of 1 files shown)


raveit65:
ohh, this PR has influences for caja.
If text is selected in pluma than the text from renaming a file or folder in caja isn't selected anymore.
I think this can drive you crazy if you work with a lot windows and workspaces, and you forgot where you did select the text.

sc0w:
@raveit65 yes, confirmed, I will investigate it

raveit65:
Same problem with mate-terminal.
I can't select anything if something in a pluma window is selected.

sc0w:
@raveit65 yes, I think it affects all the gtk applications

sc0w:
@raveit65 please test again

the behavior of this PR with the latest changes:

the selection disappears if the window lost the focus and another gtk application do one selection, and shows the selection again if the window gets the focus

raveit65:
Semms to be Ok now.
I tested it again with 3 open windows of pluma, caja and mate-terminal and i use
```
[rave@mother ~]$ gsettings get org.mate.Marco.general auto-raise
true
```
If i select text in pluma it's possible to select text in mate-terminal and CajaCanvasItem text is selected too when renaming a file in caja.
After pluma window gets focus again the text is selected again in pluma.

If you give mate-terminal or caja focus again at this point here the text isn't select any more.
But this seems to be normal behaviour of this applications and isn't related to this PR. 

With previous version of PR i noticed also a different behaviour with coping text inside mate-terminal, with using **middle-mouse-button-for-inserting** and **Copy-marked-text-to-clipboard** is enabled .

Without the PR the actual selected text from every application will be used for coping with middle-mouse-button in mate-terminal.
If you select another text from another application this text will load in memory.

With current version of PR the selected text from pluma window will be kept always in memory for using coping with middle-mouse-button in mate-terminal, no matter if you select a text in ie. firefox.

Well this behaviour sound logic for me but can be a annoying problem if you forgot a open pluma window with selected text somewhere.
So, this needs to be discussed.
@monsta @flexiondotorg @lukefromdc @XRevan86  ^^^^

PS: Hope i wrote it understandable :-)

sc0w:
> With previous version of PR i noticed also a different behaviour with coping text inside mate-terminal, with using middle-mouse-button-for-inserting and Copy-marked-text-to-clipboard is enabled .

yes, confirmed, it can be handled with something like this:

```
primtxt = gtk_clipboard_wait_for_text (gtk_clipboard_get (GDK_SELECTION_PRIMARY));

gtk_clipboard_set_text (gtk_clipboard_get (GDK_SELECTION_PRIMARY), primtxt, strlen (primtxt));
```

I will investigate it, and I will close this PR by the moment

Issue #261, opened 2017-11-15
new pluma document doesn't show in the same window if it exists in the same workspace


sc0w:
#### Expected behaviour

open new pluma document in the same window if it exists in the same workspace

#### Actual behaviour

it doesn't open in the same window

#### Steps to reproduce the behaviour

the easy way:

- open an existing file using terminal with (caja-open-terminal) **right click on folder -> open in terminal**  ( $ pluma filename )

- go in panel to `applications -> accesories -> pluma`

it open new pluma window instead the previous window

this is wrong, but it is very wrong the behavior of notebook in the new window:

go to the window with the "unsaved document" and see the behavior of mouse over the tab:

right click: it doesn't show the popup menu

make new tab in that window, and try to move it with the mouse: it doesn't work

all is wrong

#### MATE general version

1.18.0

#### Package version

it happens with pluma 1.18.2 and 1.8.1 (yes, very old version) with gtk3 (gtk 3.22.24), with 1.8.1 in gtk2 it works as expected

#### Linux Distribution

debian buster/testing

raveit65:
Hmm, i followed your steps to reproduce but here pluma opening from menu gives me always a new tab in existing pluma window.
I am using pluma from master here.

raveit65:
Same with opening pluma from terminal.
https://www.dropbox.com/s/s60jtl0btcat1yk/pluma.ogv?dl=0

sc0w:
@raveit65 I edit my first comment, try again, but open with terminal the first file: $ pluma filename

raveit65:
I did open the fist window from terminal and second from menu.
The result is the same . I have one pluma window with 2 tabs.

raveit65:
Same with 1.18.3
[rave@f26 ~]$ rpm -qa pluma
pluma-1.18.3-1.fc26.x86_64

sc0w:
see the video:

[out-51.ogv.zip](https://github.com/mate-desktop/pluma/files/1489335/out-51.ogv.zip)

weird, debian issue?

I can reproduce it too in debian stable/stretch vm with pluma 1.16.1

@lukefromdc @monsta can you reproduce this issue?

raveit65:
I did the same steps like your in this video. i can't reproduce it.
https://www.dropbox.com/s/f3f6rn43mw3s7ow/pluma-2.ogv?dl=0

sc0w:
more info: I can't reproduce the issue if I run terminal in panel with the launcher

[...]

PR #262, opened 2017-11-15
make sure pluma open in the same window of the workspace

sc0w:
there is only one screen, this code can be removed

Fixes https://github.com/mate-desktop/pluma/issues/261

### DIFF
--- a/pluma/pluma-app.c
+++ b/pluma/pluma-app.c
@@ -635,30 +635,11 @@ is_in_viewport (PlumaWindow  *window,
 		gint          viewport_x,
 		gint          viewport_y)
 {
-	GdkScreen *s;
-	GdkDisplay *display;
 	GdkWindow *gdkwindow;
-	const gchar *cur_name;
-	const gchar *name;
-	gint cur_n;
-	gint n;
 	gint ws;
 	gint sc_width, sc_height;
 	gint x, y, width, height;
 	gint vp_x, vp_y;
-
-	/* Check for screen and display match */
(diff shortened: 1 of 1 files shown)


raveit65:
Well, i really like to know why the fix is only needed for debian based distros, please read report.

lukefromdc:
I don't have any other distros to test on, nor the bandwidth to download distros for testing

monsta:
>  Btw. commit don't applies with 1.18.3

Probably due to `gdk_x11_screen_get_screen_number` function call - it's there because of deprecation fixes in git master.

sc0w:
> Btw. commit don't applies with 1.18.3

This is the adaptation for 1.18: https://github.com/mate-desktop/pluma/pull/264

monsta:
BTW, this change removes the comparing of `GdkDisplay` names too... should it be so?

monsta:
In the meanwhile, #264 works for me in Ubuntu 17.10.

sc0w:
the screen number is always the same, and that comparison can be removed safely

cur_n `0`
n `0`

@monsta 

> BTW, this change removes the comparing of GdkDisplay names too... should it be so?

yes, it works for that reason

here, in my tests:

cur_name `:0`
name `:0.0`

different values, so, return FALSE, and it makes the new window
```
	if (strcmp (cur_name, name) != 0 || cur_n != n)
		return FALSE;
```

monsta:
Ah I see why it's not reproducible everywhere... in some systems these names could be actually the same. Ok, it's good to go then.

Issue #256, opened 2017-10-18
pluma closes tab unexpectedly with middle click while the right click is holding pressed


sc0w:
how to reproduce:

in the red area:

![image4493](https://cloud.githubusercontent.com/assets/7734191/26037443/ce61d162-38f2-11e7-9616-8cc59ac650ce.png)

It needs to do **quickly** to reproduce:

1. holding pressed right click

2. middle click while the right click is pressed

3. release the buttons

(pluma 1.18.2 and git master)

raveit65:
Ok, after i found out to right-click in headers of the notebook i could reproduce the issue.

monsta:
I have a feeling I already saw such a report, and it was fixed... O_o

raveit65:
Your feeling is OK :-)
https://github.com/mate-desktop/pluma/pull/229
But the new PR is for right-clich in notebook header.
Old one was about right-click in document itself.

sc0w:
@monsta 

the "middle click implementation" to close tabs give us in the past some issues:

https://github.com/mate-desktop/pluma/issues/250 https://github.com/mate-desktop/pluma/issues/228 and https://github.com/mate-desktop/pluma/issues/215

and I hope this is the fourth and last issue

at this time, to reproduce, it needs to do **quickly** holding down [right+middle] click (in this order)

@raveit65 I can reproduce it in the document too, in the red area I described in the first comment

raveit65:
> the "middle click implementation" to close tabs give us in the past some issues:
> #250 #228 and #215
> and I hope this is the fourth and last issue

@sc0w 
But in which branch or git history is the culprit commit?
Btw. Do we need all fourth new commits for a new 1.18 point release?
Or does the issue only exist in master?

sc0w:
@raveit65 I can confirm, the commit https://github.com/mate-desktop/pluma/commit/9d1bf65037d56ced8c3d4396902e83e286222914 makes the issues https://github.com/mate-desktop/pluma/issues/228 https://github.com/mate-desktop/pluma/issues/215 and this issue https://github.com/mate-desktop/pluma/issues/256

The issue https://github.com/mate-desktop/pluma/issues/250 is not related to the middle click implementation

> Do we need all fourth new commits for a new 1.18 point release?

Three commits, the issue https://github.com/mate-desktop/pluma/issues/215 was already fixed in 1.18 branch

[...]

PR #263, opened 2017-11-20
pluma-view.c: use one static variable instead two to do the same job

sc0w:
This is the improvement of this two commits:

https://github.com/mate-desktop/pluma/commit/4df44887f4b7547fc9b31d1cb3f61688579fe39b
https://github.com/mate-desktop/pluma/commit/ec438f3d262ef178a898612211335d3c9fa2f4ee

The tests: make sure you can't reproduce the issues #250 and #256

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -57,8 +57,7 @@
 #define PLUMA_VIEW_GET_PRIVATE(object)(G_TYPE_INSTANCE_GET_PRIVATE ((object), PLUMA_TYPE_VIEW, PlumaViewPrivate))
 
 /* Local variables */
-static gboolean middledown = FALSE;
-static gboolean rightdown = FALSE;
+static gboolean middle_or_right_down = FALSE;
 
 typedef enum
 {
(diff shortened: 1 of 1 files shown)


pluma/pluma-view.c
```diff
@@ -2034,13 +2032,17 @@ static gboolean
 pluma_view_button_release_event (GtkWidget *widget, GdkEventButton *event)
 {
 	if (event->button == 2)
-		middledown = FALSE;
-	else if (event->button == 3)
-		rightdown = FALSE;
+		middle_or_right_down = FALSE;
```
raveit65: Is is OK to remove everything for event->button == 3 here ?

sc0w: yes, never works the button 3 here (middle click), I did instead the function pluma_view_populate_popup

you can test that, it fixes a little issue:

1- select text

2- right button (appears the menu)

3- middle button (dissapears the menu)

4- middle button again

- the expected, with this PR: paste the text

- without this PR: nothing happens, it needs one more middle click

raveit65: Ok, got it

monsta: Confirmed the issue and the fix.

PR #275, opened 2018-01-28
require GTK+ 3.22 and GLib 2.50

monsta:


monsta:
Oops, somehow I missed that one. Ok, corrected and force-pushed.

lukefromdc:
Updated PR builds and runs fine here on Debian Unstable, same as before.

sc0w:
there is no more checks, and the build is ok

PR #276, opened 2018-02-04
Switch tabs by mouse wheel

### LINKED ISSUE #271: [Feature Request] Allow scrolling on notebook tabs to change the active page
In Linux Mint in Xed has been added "Allow scrolling on notebook tabs to change the active page", but this does not work in Fedora 27. Can you add a similar functionality to Pluma?
xed

likorisd:
Should fix #271

### DIFF
--- a/pluma/pluma-window.c
+++ b/pluma/pluma-window.c
@@ -36,6 +36,7 @@
 #include <sys/types.h>
 #include <string.h>
 
+#include <gdk/gdk.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
 #include <gtk/gtk.h>
@@ -3489,6 +3490,23 @@ notebook_button_press_event (GtkNotebook    *notebook,
 }
 
 static gboolean
+notebook_scroll_event (GtkNotebook    *notebook,
+                       GdkEventScroll *event,
+                       PlumaWindow    *window)
+{
+	if (event->direction == GDK_SCROLL_UP || event->direction == GDK_SCROLL_LEFT)
+	{
+		gtk_notebook_prev_page (notebook);
+	}
+	else if (event->direction == GDK_SCROLL_DOWN || event->direction == GDK_SCROLL_RIGHT)
+	{
+		gtk_notebook_next_page (notebook);
+	}
+	
+	return FALSE;
+}
+
+static gboolean
 notebook_popup_menu (GtkNotebook *notebook,
 		     PlumaWindow *window)
 {
@@ -3854,6 +3872,10 @@ connect_notebook_signals (PlumaWindow *window,
 			  "popup-menu",
 			  G_CALLBACK (notebook_popup_menu),
 			  window);
+	g_signal_connect (notebook,
+			  "scroll-event",
+			  G_CALLBACK (notebook_scroll_event),
+			  window);
 }
 
 static void
@@ -3866,7 +3888,8 @@ add_notebook (PlumaWindow *window,
 	                 TRUE);
 
 	gtk_widget_show (notebook);
-
+	
+	gtk_widget_add_events (notebook, GDK_SCROLL_MASK);
 	connect_notebook_signals (window, notebook);
 }
 
--- a/pluma/pluma-notebook.c
+++ b/pluma/pluma-notebook.c
@@ -843,9 +843,11 @@ tab_label_scroll_cb (PlumaTabLabel  *tab_label,
                      GdkEventScroll *event,
                      PlumaNotebook  *notebook)
 {
-	g_signal_emit_by_name (notebook, "scroll-event", &event);
+	gboolean retval;
 	
-	return FALSE;
+	g_signal_emit_by_name (notebook, "scroll-event", &event, &retval);
+	
+	return retval;
 }
 
 static GtkWidget *
--- a/pluma/pluma-notebook.c
+++ b/pluma/pluma-notebook.c
@@ -843,11 +843,9 @@ tab_label_scroll_cb (PlumaTabLabel  *tab_label,
                      GdkEventScroll *event,
                      PlumaNotebook  *notebook)
 {
-	gboolean retval;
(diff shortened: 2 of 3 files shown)


raveit65:
So it works only if the mouse is over free space or arrows in notebook header?
First i thought the PR didn't work as i tested it with more than 2 tabs :-)

pluma/pluma-window.c
```diff
@@ -3854,6 +3872,10 @@ connect_notebook_signals (PlumaWindow *window,
 			  "popup-menu",
 			  G_CALLBACK (notebook_popup_menu),
 			  window);
+	g_signal_connect (notebook,
```
raveit65: Can you please fix intends?

likorisd: Whoops, sorry. Sure...

raveit65:
commits should be squashed local before merging.

likorisd:
Like this? (Sorry, I'm still new to using git versioning...)

raveit65:
This was more a hint for the person who will merge the PR.
Well, commit looks good now, but commit message is weird now ?
See https://patch-diff.githubusercontent.com/raw/mate-desktop/pluma/pull/276.patch

I am using 'git rebase -i last-commit-id-before-you-want-to-edit ' for editing commits.
i = interactive
This will open an interactive vi editor in terminal.
... it' easy.
Than use 'git push -f' to update your repo with your local changes.
The PR will be updated automatic after this ;-)

vkareh:
@likorisd - some preliminary testing tells me that the best place to capture the scroll event (within the tabs) is the `ebox` widget (inside `pluma-tab-label.c` - `pluma_tab_label_init()`)

However, it only works if you remove the scroll-event in the Notebook widget, otherwise it ends up double-scrolling... which means that scroll only happens on the tabs, but not on the empty area... tradeoffs 😛

lukefromdc:
Scrolling only when over the tabs is a lot more natural and easily discovered than scrolling only when not over them in my opinion.

vkareh:
@lukefromdc - I fully agree. I would prefer scrolling on either to work, but I'd be perfectly happy with scrolling on just the tabs themselves.

raveit65:
> However, it only works if you remove the scroll-event in the Notebook widget, otherwise it ends up double-scrolling... which means that scroll only happens on the tabs, but not on the empty area... tradeoffs 😛

This is OK, i see no reason why scrolling should work in the empty area.
Mostly, i don't have an empty notebook header, because a lot files are open....

Yetoo1:
> This is OK, i see no reason why scrolling should work in the empty area.
This isn't taking into account the people who have few tabs open and who use a mouse to scroll. Without both areas allowing scroll to happen, they need to mouse their mouse more to get it into correct position to scroll through the tabs. Subconsciously or not, the user need to remember the area to move the cursor over to in order to scroll. If there was just a single invisible box that overlaid the tab area, this would mitigate any trade offs and give a single place to remember to put the mouse to scroll.

vkareh:
An alternative would be to draw tabs like mate-terminal does: spanning the entire width of the `GtkNotebook`, and also hide the tab bar if there's only one document open.

That would make it consistent across apps. Now, about shortcuts... 🙄

Yetoo1:
If Pluma were to have tabs span all across `GtkNotebook` (I personally find this annoying and too obtrusive) Pluma, the shortcuts need to be much more similar to nautilus rather than mate-terminal. Mate-terminal is an exception because of it being a terminal emulator and taking more keystrokes literally.

likorisd:
I passed the scroll-event from `pluma-tab-label.c` to the notebook. This seems to work with the cursor both on and off the tab label.

raveit65:
First commit message should be reworded back to origin meaning, this can be done by person who merge this PR.

raveit65:
guys, does this works for you?

vkareh:
Hmmm.... This didn't work for me :/

vkareh:
Wait, i just did it with the old patch and it didn't work either, so it's probably my fault... Man, my build is really botched today... I'll try again later tonight

vkareh:
@monsta - as for the return value of the callback. Do you think this needs to be changed, or do we merge as is?

likorisd:
Thanks for your reviews. :)

As for the return values: I tried to change them to `TRUE` but this behaves weird. It lets me scroll up (left) with the mouse pointer on top of the labels, but not down (right). Scrolling with the mouse pointer outside the labels works in both directions.

So, I'd suggest to leave the return value as `FALSE` for now.

vkareh:
@likorisd - returning `TRUE` prevents further propagation of the event. In your case you have the event signal being captured in several places. Based on what you just said, it seems as though stopping the event propagation is preventing some of the other callbacks from being fired, maybe?

I'm ok with `FALSE`, as it allows other parts of the code to read scroll events higher up the widget hierarchy.

likorisd:
I changed this though to pass the return values the same way back. So it's easier to change later, if needed.

likorisd:
Hmm, yes, maybe. I'll look into it.

likorisd:
Okay, cascading the return value as `TRUE` back to `pluma-tab-label.c` makes no sense. At least in `pluma-notebook.c` the callback should return `FALSE`. So forget my last commit...

raveit65:
I tested it again and it works for me.
@monsta @vkareh 
Is this ready to go now?

vkareh:
I just tested it again and it works well for me too. I'd say this is ready to go :+1:

monsta:
I did a quick test yesterday, seemed to work. I'll check the code just in case

monsta:
Looks good to me. Now what was the original message in the first commit...?

>  Well, commit looks good now, but commit message is weird now ?

First two commits need to be squashed into a single commit with some meaningful message, and the last two commits should be dropped as together they are a no-op.

raveit65:
rebased and merged
likorisd-dev-scrollable-tabs
Thank you

monsta:
@likorisd: please look at https://github.com/mate-desktop/pluma/pull/320.

PR #284, opened 2018-03-02
python console: config.ui: avoid GtkButton:use-stock

sc0w:
the expected: the same look and behavior like before


raveit65:
How can i execute this config-dialog for python_console?

sc0w:
I don't know, the python console doesn't work here without/with the PR

probably this dialog shows in edit -> preferences -> plugins -> python console -> preferences button

sc0w:
ok, if all is working like before, I will merge this PR

sc0w:
@monsta do you know how to show this dialog?

PR #286, opened 2018-03-02
spell plugin: pluma-spell-setup-dialog.ui: avoid deprecated:

sc0w:
avoid deprecated GtkVBox and GtkButton:use-stock

I noticed the buttons in the ui file: '`Help`', '`Cancel`', and '`OK`', never works without/with the PR

spell_with_pr


raveit65:
Indeed , there are no icons in buttons without this PR, but in glade i see them.....

PR #289, opened 2018-03-04
spell plugin: languages-dialog.ui: avoid deprecated:

sc0w:
avoid GtkButton:use-stock, GtkVBox, and GtkLabel:xpad/ypad

and now, there is no deprecations

NOTE: I noticed the three buttons are handled by code, because I have tested with/without the PR, and the GtkImages always have the stock properties


sc0w:
merged: https://github.com/mate-desktop/pluma/commit/028fb442f60a3fc7f02d5a791decc58ab577bcff

PR #295, opened 2018-03-07
pluma-time-dialog.ui plugin: avoid deprecated:

sc0w:
avoid deprecated:

GtkVBox
GtkHBox
GtkLabel:xpad/ypad
GtkImage:xalign/yalign/xpad/ypad/stock
GtkButton:use-stock

and add icon to the 'Insert' button

time1

time2

### DIFF
--- a/plugins/time/pluma-time-dialog.ui
+++ b/plugins/time/pluma-time-dialog.ui
@@ -1,96 +1,110 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.20.2 -->
 <!--*- mode: xml -*-->
 <interface>
-  <requires lib="gtk+" version="2.16"/>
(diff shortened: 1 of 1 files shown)


raveit65:
With setting orientation=vertical on dialog-vbox2 the buttons will move to the bottom in glade-preview.
```
      <object class="GtkBox" id="dialog-vbox2">
        <property name="visible">True</property>
        <property name="can_focus">False</property>
        <property name="orientation">vertical</property>
        <property name="spacing">2</property>
```
For some reasons the default *horizontal orientation* does work in reality too,
I saw that in a lot of UI files.

sc0w:
yes, confirmed, but the final result is the same

sc0w:
with latest changes:

I changed the orientation to vertical in dialog-vbox2 and I fixed the focus using the <tab> key in all the items

raveit65:
All fine, ready to go.

PR #296, opened 2018-03-07
Fix make distcheck

info-cppsp:
A file was missing from translations.


raveit65:
Thanks for fixing that.

PR #299, opened 2018-03-10
sort.ui plugin: avoid deprecated:

sc0w:
avoid deprecated:

GtkVBox
GtkHBox
GtkLabel:xpad/ypad
GtkImage:xpad/ypad/xalign/yalign/stock
GtkButton:use-stock

sort1

sort2


lukefromdc:
Daylight Savings time messing up timestamps

PR #303, opened 2018-03-30
pluma-time-plugin.c: avoid 'gtk_dialog_new_with_buttons' with stock id

sc0w:
I have no idea how to test this PR

### DIFF
--- a/plugins/time/pluma-time-plugin.c
+++ b/plugins/time/pluma-time-plugin.c
@@ -916,11 +916,11 @@ get_choose_format_dialog (GtkWindow                 *parent,
 	{
 		GtkWidget *err_dialog;
 
-		err_dialog = gtk_dialog_new_with_buttons (NULL,
-			parent,
-			GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT,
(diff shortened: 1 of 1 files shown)


raveit65:
Looks like that in the time-plugin-dialog from preferences an icon is missing for the close button.
As the button is translated i looks like an old issue and not related to PR.

sc0w:
aha, finally I found the dialog:

it happens with error to load the .ui file:

time1

error_time_dialog

The PR seems to work fine here

@raveit65 

> Looks like that in the time-plugin-dialog from preferences an icon is missing for the close button.
> As the button is translated it looks like an old issue and not related to PR.

yes, by the moment, I don't know how to fix that

raveit65:
@sc0w 
> it happens with error to load the .ui file:

Local issue or a blocker?
I don't see this problem here.

sc0w:
@raveit65 no, there is no problems

I mean: this PR affects this dialog (the "ok" button):

error_time_dialog

to show the dialog, you need to modify the file pluma-time-dialog.ui to force the error dialog

raveit65:
Ahh, Ok
let's merge it.

PR #306, opened 2018-04-14
avoid 'pluma_file_chooser_dialog_new' with stock ids

sc0w:


raveit65:
All icons are there in pluma filechooser dialog.

PR #320, opened 2018-05-21
Fix: Pluma crashes with mouse wheel over a tab

sc0w:
Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=898637

I can't reproduce the debian bug, but reporter says this PR fixes the problem

In the tests, make sure the PR https://github.com/mate-desktop/pluma/pull/276 is working as before

### DIFF
--- a/pluma/pluma-notebook.c
+++ b/pluma/pluma-notebook.c
@@ -864,16 +864,6 @@ close_button_clicked_cb (PlumaTabLabel *tab_label, PlumaNotebook *notebook)
 	g_signal_emit (notebook, signals[TAB_CLOSE_REQUEST], 0, tab);
 }
 
-static gboolean
-tab_label_scroll_cb (PlumaTabLabel  *tab_label,
-                     GdkEventScroll *event,
-                     PlumaNotebook  *notebook)
-{
-	g_signal_emit_by_name (notebook, "scroll-event", &event);
-	
-	return FALSE;
-}
-
 static GtkWidget *
 create_tab_label (PlumaNotebook *nb,
 		  PlumaTab      *tab)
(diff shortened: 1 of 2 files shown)


raveit65:
Weird, i can't reproduce this issue. Is this confirmed by anyone else than reporter?

sc0w:
I don't know, I can't reproduce the crash, but seems the code removed here is useless, because https://github.com/mate-desktop/pluma/pull/276 is working exactly as before

monsta:
Hmm, that code had been added in https://github.com/mate-desktop/pluma/pull/276/commits/3ef5682dfdec9ef27d3b8c1e43bfaf0c91138538 to allow scrolling when mouse pointer is both over the tab and over the empty space to the right (if you don't have many tabs open). Need to check it... and I can't reproduce it with git master too.

monsta:
BTW, thanks to the bug reporter for building, testing and getting the backtrace. If only all bug reporters were like that...

raveit65:
Hmm, why will something approved without the issue is confirmed by a second person?

sc0w:
confirmed or not, this is a code improvement, we don't need useless code :)

raveit65:
I am missing a confirmation by debian maintainer in upstream report.
I saw too much weirdness with GtkNotebook and pointer events in pluma today, see other new report.
Hopefully this causes not a regression because of a unconfirmed issue, possible a local problem by a debian user. 
If yes i will say i told you so ;)

raveit65:
btw. go ahead if you think this is save......

sc0w:
Finally, I can reproduce the crash!

with gtkinspector pluma crashes, and I confirm this PR fixes the issue

see the video:

[out-59.ogv.zip](https://github.com/mate-desktop/pluma/files/2027010/out-59.ogv.zip)

PR #318, opened 2018-05-13
pluma-notebook.c: avoid deprecated 'gdk_pointer_ ...' functions

sc0w:
avoid deprecated:

gdk_pointer_is_grabbed
gdk_pointer_grab
gdk_pointer_ungrab

PR #326, opened 2018-05-24
pluma-notebook: pass event to gdk_seat_grab

monsta:
follow-up to d6d1cdb0e241ee3a07200ad0561a8105b6243c2b (see https://github.com/mate-desktop/pluma/pull/318#discussion_r190516342)

### DIFF
--- a/pluma/pluma-notebook.c
+++ b/pluma/pluma-notebook.c
@@ -394,7 +394,7 @@ pluma_notebook_reorder_tab (PlumaNotebook *src,
 
 static void
 drag_start (PlumaNotebook *notebook,
-	    guint32        time)
+            GdkEvent      *event)
 {
 	GdkSeat    *seat;
 	GdkDevice  *device;
@@ -424,7 +424,7 @@ drag_start (PlumaNotebook *notebook,
 			       GDK_SEAT_CAPABILITY_POINTER,
 			       FALSE,
 			       cursor,
-			       NULL,
+			       event,
 			       NULL,
 			       NULL);
 	}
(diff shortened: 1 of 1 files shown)


lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/pluma/commit/72edb5de006251f6d561291ccf230d685bfb5f16

monsta:
I would suggest using "Rebase and merge" button.

rebase-and-merge-button

You need to press the down arrow there and select the option.

It marks the PR as merged instead of closed, and also adds a link to the PR to the merged commit. The link is only visible on Github, so it doesn't clutter the commit history when you use git from the command line.

lukefromdc:
I did some tests in  https://github.com/lukefromdc/test/network and found that "Rebase and merge" will work and give the same results as --ff-only with a single commit. Unfortunately, "Squash and merge" with a chain of commits puts the branch name in as the top commit message. Thus a squash has to be pushed first to the PR branch, and if that is rebased locally (one command) it can then merge and automatically show the PR as merged.

raveit65:
In most cases squash + merge isn't usefull here as it squash all commits.
So those action od edititing a commit i do always local.
"Rebase and merge" is nice to do at website because of the added link here at github.

lukefromdc:
Got a bit of a mess with the default merge button (one commit not needing a rebase) on mate-calc:
https://github.com/mate-desktop/mate-calc/pull/53#issuecomment-392152772

monsta:
I recall you can edit the commit message when using "Squash and merge" button... By default, it also puts the PR number in the message, which isn't needed as the commit will have a link to PR anyway. I remember I was able to edit it and remove the number before merging.

Yes, "Squash and merge" is only useful if you want to squash all the commits. If you need to squash only a few and leave others as is, it has to be done manually.

Finally, the default merge button always adds a new merge commit, which doesn't look good...

PR #340, opened 2018-07-09
pluma-view: Fix: Bold and Italic fonts don't work

sc0w:
culprit commit: https://github.com/mate-desktop/pluma/commit/310bfc249aac3dd80511d3dade14aa06f3d9769a (my fault)

Since that commit bold and italic fonts don't work

This is the fix

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -647,32 +647,43 @@ pluma_view_scroll_to_cursor (PlumaView *view)
 }
 
 static void
-pluma_override_font (GtkWidget   *widget,
-		     const gchar *font)
+pluma_override_font (GtkWidget            *widget,
+		     PangoFontDescription *font)
 {
-	gchar          *css;
 	GtkCssProvider *provider;
-	gchar          *tempsize;
+	gchar          *css;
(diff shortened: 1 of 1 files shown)


lukefromdc:
We also get this new warning running from terminal from this commit, not present in master:

```
(pluma:18525): Gtk-WARNING **: 17:47:24.192: Theme parsing error: <data>:1:17: not a number

(pluma:18525): Gtk-WARNING **: 17:47:24.192: Theme parsing error: <data>:1:27: Using Pango syntax for the font: style property is deprecated; please use CSS syntax


```

sc0w:
warnings fixed with latest changes

@lukefromdc please test again

lukefromdc:
OK, I con confirm that a build with the latest change no longer throws the runtime warnings I got last time, and bold and italic fonts work

sc0w:
more changes, now the system font works if it is in px ("Monospace 10px" for example)

lukefromdc:
I just retested and can confirm that this makes system bold and italic monospace fonts work, also makes same fonts selected from within Pluma itself work, and does not throw any warnings. Seems to be ready to go

Issue #234, opened 2017-06-12
Context menus use a monospace font.


flexiondotorg:
#### Expected behaviour

Right clicking in the text area of Pluma should display a context menu that using the system font.

#### Actual behaviour

Right clicking in the text area of Pluma displays a context menu that uses the "Editor font" configured in the Pluma Preferences.

#### Steps to reproduce the behaviour

See *Actual behaviour* above.

#### MATE general version

  * MATE Desktop 1.18

#### Package version

  * pluma 1.18.2-0ubuntu1

#### Linux Distribution

  * Ubuntu MATE 17.10 daily

monsta:
We should "thank" GNOME devs for that.
https://bugzilla.gnome.org/697127

sc0w:
I need to know exactly how to get the "system font"

Here, in this gsettings key?

`/org/mate/desktop/interface/font-name`

alexarnaud:
Le 01/06/2018 à 18:00, ZenWalker a écrit :
> Here, in this gsettings key?
> 
> |/org/mate/desktop/interface/font-name|

Right, this is this one.

Best regards,
Alex.

biscottaro:
This problem is still present on Pluma 1.26.0 on Debian testing as of 02/06/22. Should i open a new report? I can't reopen this one.

lukefromdc:
I cannot duplicate this with Pluma from git master over Debian Unstable

PR #343, opened 2018-07-18
pluma-view: Fix font in context menu

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/234


vkareh:
I have a valgrind report here: https://paste.ubuntu.com/p/qrrCtk74Rp (search for pluma-view.c in there)

sc0w:
@vkareh thanks for your tests, now is fine with latest changes?

lukefromdc:
Behavior is fine here, same as on the first try. Still have to open a new Pluma window to apply a change of system font to the context menu, so that didn't come from the const issue. 

Taking that out and keeping the g_free statements works here with no ugly surprises though. Of all new variables set, it looks like `desc` is returned, the returned value freed at the end of `pluma_view_set_font ` and everything else freed so no reason for any more memory leaks here.

muktupavels:
First call to `pluma_view_set_font`:
- textview { css; } .context-menu { css; }

Next call to `pluma_view_set_font`:
- textview { css; } .context-menu { css; } textview { css; } .context-menu { css; }

And next one:
- textview { css; } .context-menu { css; } textview { css; } .context-menu { css; } textview { css; } .context-menu { css; }

... you can continue forever... One call generates css something like this:

> textview { font-family: X; font-weight: %X; font-style: normal; font-size: Xpt; } .context-menu { font-family: X; font-weight: %X; font-style: normal; font-size: Xpt; } 

Looks damn good, right? Right now I have gedit open with 3 windows and close to ~40 tabs (GeditViews or in your case it would be PlumaView). That is 40 * 169 bytes (as minumum). Thats ~6.6 kb usage. My pc uptime is 37 days and I probably have closed gedit exactly zero times... But I do open and close tabs every day. So at this point that function might have been called 1000+ times. So that like ~165kb of css usage.

And also GTK has to parse that css... Bigger it grows slower it will happen.

PR #243, opened 2017-08-08
Adds support for ctrl+tab and ctrl+shift+tab

snaiperskaya96:

PR #346, opened 2018-07-25
add the abbility to switch tabs using [ctrl+tab] and [ctrl+shift+tab]

sc0w:
If true the gsettings key "ctrl-tab-switch-tabs" into "org.mate.pluma"

Closes https://github.com/mate-desktop/pluma/issues/211

We have since a long time ago the PRs https://github.com/mate-desktop/pluma/pull/243 and https://github.com/mate-desktop/pluma/pull/242 with no news, and I like this feature, so, I decide to open this alternative PR.

We have the choice with the gsettings key.

### DIFF
--- a/data/org.mate.pluma.gschema.xml.in
+++ b/data/org.mate.pluma.gschema.xml.in
@@ -11,6 +11,11 @@
       <summary>Editor Font</summary>
       <description>A custom font that will be used for the editing area. This will only take effect if the "Use Default Font" option is turned off.</description>
     </key>
+    <key name="ctrl-tab-switch-tabs" type="b">
+      <default>false</default>
+      <summary>Switch tabs with [ctrl] + [tab]</summary>
+      <description>If true, it enables the ability to switch tabs using [ctrl + tab] and [ctrl + shift + tab].</description>
+    </key>
     <key name="color-scheme" type="s">
       <default>'tango'</default>
       <summary>Style Scheme</summary>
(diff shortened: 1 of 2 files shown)


vkareh:
This is awesome! I really want this feature! I'll test it tomorrow when I get to work. Just one thing. Is there a reason to hide it behind the gsettings flag? I feel this should be a default key binding and it shouldn't conflict with anything else, I think, so we could just as well have it always on

lukefromdc:
We need to make sure all the keyboard shortcuts used by those who have issues using a mouse work with this, I really don't know them all myself. See the comments made in both https://github.com/mate-desktop/pluma/pull/242 https://github.com/mate-desktop/pluma/pull/243 on these issues

sc0w:
@vkareh with that keys we break the standard behavior to change between items in all the gtk applications, so, we need the gsettings key.

alexarnaud:
Le 25/07/2018 à 05:11, lukefromdc a écrit :
> We need to make sure all the keyboard shortcuts used by those who have 
> issues using a mouse work with this, I really don't know them all 
> myself. See the comments made in both #242 
> <https://github.com/mate-desktop/pluma/pull/242> #243 
> <https://github.com/mate-desktop/pluma/pull/243> on these issues

IMO, as it's disabled by default I don't think it could have impact on 
accessibility. I personally think ctrl+tab is a must have for Pluma.

Best regards,
Alex.

PR #350, opened 2018-07-28
 add the option to hide the frist tab if there is only one tab 

sc0w:
In the tests, make sure the key `show-tabs-with-side-pane` using the side pane is working as expected:

`show-tabs-with-side-pane` _false_ and `show-first-tab` _true_: all the tabs are hidden.
`
show-tabs-with-side-pane` _true_ and `show-first-tab` _false_: it hides the first tab (if there is only one tab)


lukefromdc:
It was not `show-tabs-with-side-pane` that needed to add or remove a tab to apply changes but rather  the new `show-first-tab` option

sc0w:
@lukefromdc the behavior is the same, you can press F9 and the `show-first-tab` option works

Both options needs: press F9, make new tab or remove existing tab

lukefromdc:
I hadn't noticed that with show-tabs-with-side-pane simply because that was the first time I have ever used or seen it. Ideally we'd have a current, up-to-date list of all the features in each app (the help is out of date) but we may not have enough team resources to do that at this point.

sc0w:
The gsettings key sets whether to show the tab for the notebook or not if we have only one tab. If there is more tabs, the first tab isn't affected, obviously.

It works with / without the sidebar.

The key is for users to like open frequently one unique file, and the tab for notebook isn't needed.

I will change the key name later.

sc0w:
key renamed

btw, we have already this feature but seems never works, see this comment in code:

https://github.com/mate-desktop/pluma/blob/master/pluma/pluma-notebook.c#L782

PR #351, opened 2018-08-04
Font refactoring

lukefromdc:
For the Pluma  context menus, use the "intial" cssvalue to return context menus to system font whatever they may be. Changes are applied immediately, requiring no new signal to be connected and removing the need for the  get_system_font function.

Revert the previous context menu work to get back to the starting point. Afterwards, clean things up a bit: create and add the static cssprovider in a single block, write to it afterwards. Shuffle some variable declarations to get a consistant block as when they are freed at the end. Document all this with comments in code.

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -646,33 +646,12 @@ pluma_view_scroll_to_cursor (PlumaView *view)
 				      0.0);
 }
 
-static PangoFontDescription* get_system_font (void)
-{
-    PangoFontDescription *desc = NULL;
-    GSettings *settings;
-    char *name;
-
-    settings = g_settings_new ("org.mate.interface");
-    name = g_settings_get_string (settings, "font-name");
-
-    if (name) {
-    	desc = pango_font_description_from_string (name);
-    	g_free (name);
-    }
-
-    g_object_unref (settings);
-
-    return desc;
-}
-
 static void
-pluma_override_font (const gchar          *item,
-		     GtkWidget            *widget,
+pluma_override_font (GtkWidget            *widget,
 		     PangoFontDescription *font)
 {
 	static gboolean provider_added = FALSE;
 	GtkCssProvider *provider;
-	gchar          *prov_str;
 	gchar          *css;
 	gchar          *family;
 	gchar          *weight;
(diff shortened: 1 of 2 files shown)


lukefromdc:
This is less code to run, and maybe 1kb less in RAM according to Albert's comments about the size of a cssprovider as used in mate-applets/weather applet.

lukefromdc:
I found that I also could not apply a bold or italic font as the application font anywhere else other than the Pluma context menu(with current master) so we have a bigger problem there. I will update the text on this to reflect the use of a single style provider in master.

This is a refactoring effort so there is no reason to worry about doing it fast, this is about doing it as well as possible instead.

sc0w:
Yes, the bold / italic font in context menu, and in all menus, must work everywhere. We have another issue, and I am not sure what mate component is the culprit.

It works as expected in mate 1.8.1 (gtk 2)

lukefromdc:
On the road now, I will open an issue against mate-settings-daemon as the problem
is likely there. The mate-control-center gui successfully stores the font style, or we
would not be able to pick it up in master. The problem is that that value is not being
applied anywhere except where it is localy read and (re)applied.

sc0w:
It works in firefox

lukefromdc:
On further testing, I was also unable to change the application font to a bold or italic style in Cinnamon or GNOME. In Cinnamon behavior was the same as in MATE: font size could be changed but not font style. In GNOME changing the application font was completely broken for some reason.

Attempts to set bold or italic application fonts get as far as being stored in the dconf database but are then ignored. I don't have KDE installed so can't test font handling there, but I did find that QT and GTK2 apps also could not apply a bold or italic application font so GTK is probably not to blame.

I now wonder if the maintainers of fontconfig or some other underlying library decided to block setting an italic or bold font as the default. Remember that a bold default font cannot be set any heavier except by a numerical value rather than a style name.

Thus, the issue here with bold or italic fonts in context menus may have nothing to do with MATE at all, and should not affect this PR.  Anything we can apply as the application font works in the Pluma context menu here, these being changes to font-size or font-family.

sc0w:
The fix for pluma: https://github.com/mate-desktop/pluma/pull/352

lukefromdc:
I do not support trying to enable bold and italic fonts on a per-application basis as we accidently did here on Pluma for just the context menus as master currently stands. I see an incredible amount of headache potential in trying to find all the missed labels, the bug reports that would come in when someone finds they can set a bold or italic system font-but it works only on MATE apps, only some places in them, and maybe not at all in some of them.   

We need to find the underlying issue, not patch around it and then have to revert all the kludges when the real problem gets fixed elsewhere. Since this is not limited to MATE, we need to find out where it is, then report it to whoever maintains the offending package. If they won't fix it, (say because they think nobody should use a bold or italic font for the system font), then we can fork the offending package or add an extension to mate-settings daemon and/or mate-desktop that has that effect.

This PR can sit for now, until the font issue is found elsewhere and it can then be properly tested.

lukefromdc:
Rebased after https://github.com/mate-desktop/pluma/commit/eb7a39bd311d89bf0e798f9d4b2cda80596e9217 for the convenience of other testers

lukefromdc:
I am walking away from this PR, had enough

lukefromdc:
Branch reposted in case anyone else wants to play with this or test it.

PR #352, opened 2018-08-06
pluma-view: Fix bold and italic fonts in menus

sc0w:
how to test:

go to `system -> preferences -> look and feel -> appearance -> fonts ->` change the `application font` to a `bold` or `itailc` font:

screenshot at 2018-08-06 02-31-29

I think this is an accessibility fix

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -667,15 +667,15 @@ static PangoFontDescription* get_system_font (void)
 }
 
 static void
-contextmenu_font_changed_cb (GSettings *settings,
-			     gchar     *key,
-			     gpointer   user_data)
+system_font_changed_cb (GSettings *settings,
+			gchar     *key,
+			gpointer   user_data)
 {
 	PangoFontDescription *sys_font_desc = NULL;
 	sys_font_desc = get_system_font ();
 	if (sys_font_desc)
 	{
-		pluma_override_font (".context-menu", NULL, sys_font_desc);
+		pluma_override_font ("label", NULL, sys_font_desc);
 		pango_font_description_free (sys_font_desc);
 	}
 }
@@ -714,9 +714,9 @@ pluma_override_font (const gchar          *item,
 
 	prov_str = gtk_css_provider_to_string (provider);
 
-	if (g_str_has_prefix (prov_str, "textview") && g_str_has_prefix (item, ".context-menu"))
+	if (g_str_has_prefix (prov_str, "textview") && g_str_has_prefix (item, "label"))
 	{
-		if (strstr (prov_str, ".context-menu"))
+		if (strstr (prov_str, "label"))
 		{
 			g_strdelimit (prov_str, "}", '\0');
 			gchar *prov_new_str = g_strdup_printf ("%s}", prov_str);
@@ -737,7 +737,7 @@ pluma_override_font (const gchar          *item,
 		settings = g_settings_new ("org.mate.interface");
 		g_signal_connect (settings,
 				  "changed::" "font-name",
-				  G_CALLBACK (contextmenu_font_changed_cb), NULL);
+				  G_CALLBACK (system_font_changed_cb), NULL);
 
 		gtk_style_context_add_provider_for_screen (gtk_widget_get_screen (widget),
 							   GTK_STYLE_PROVIDER (provider),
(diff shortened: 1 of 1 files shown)


lukefromdc:
One more point: we can do nothing about Firefox by fixing this in MATE.

sc0w:
> One more point: we can do nothing about Firefox by fixing this in MATE.

In Firefox the fonts are applied as expected with bold or italic fonts.

I think this fix is good for persons with visual problems.

sc0w:
It works here in debian testing with `firefox-esr 52.9.0`

screenshot at 2018-08-07 01-14-41

... and **it works** in aMule 2.3.2, KeePassX 2.0.3, Inkscape 0.92.3, Xsane 0.999, Jeex 12.0.4, Poedit 2.0.9, grsync 1.2.6, recordMyDesktop 0.3.8.1, gparted 0.30.0, Thunar 1.6.15, and probably more applications...

lukefromdc:
I don't know what you have different.  I have Inkscape, does not work there. Have Xsane, does not work there, Same in the recordemydesktop GUI, but worked fully in Gparted. Wondered it it was permissions but turned out not to be, Tried Inkscape as root, did not apply a bold or an italic font. Most of the other programs mentioned I do not have.

Firefox ESR(Torbrowser) was an interesting case: a bold font could be applied, but not an italic font! Applying a bold italic font gave a bold font only. I really don't know much about fontconfig(which most programs are supposed to read for font info in X), but maybe something in there is not working or maybe a feature was removed? Also wonder if Firefox ESR is reading gsettings directly and filtering out italic?

As for this PR, I will not myself merge a single app fix for this, only a system level fix, though I won't hold the PR's with "changes requested" either.  If others here really want to go with application level fixes, consider checking the system font for "bold" or Italic" and not running override_font (or returning from it) when these are not present. 

Leaving this for other team members to decide, since our opinons on this are opposite.  My rationale is this, assuming the fontconfig devs did this on purpose: If the system font is already a bold font, any calls to use a bold font elsewhere do not make the font further boldened. Same for italics. Thus, for accessability purposes the correct fix (barring just using a larger font), would be for someone to provide a font which is very heavy (like a bold font) by default, and has a further bold version, a further bold/italic version, and an italic version fo cover all the cases.

lukefromdc:
Rebased after https://github.com/mate-desktop/pluma/commit/eb7a39bd311d89bf0e798f9d4b2cda80596e9217 for the convenience of other testers

raveit65:
About what gsettings key are you talking?
Btw. which key does not work with all (most) applications?

raveit65:
All without this or the other PR.
In rhel7.5 VM with MATE-1.16 (gtk2)
rhel-7 5-mate-1 16-gtk2-sans-bold-10
Here using
```
[rave@qemu-rhel7 ~]$ gsettings get org.mate.interface font-name
'Sans bold 10'
```
works fine with all gk2 applications from MATE and firefox-52.8.0 (gtk2).
But not with Mate gtk3 apps like mate-terminal, engrampa or gnome csd applications like evince.

In my main box with all from master and gtk+-3.23.1 it works out of box with thunderbird.
thunderbird-bold-fonts

So what does thunderbird better than all other applications?
For this looks like there is more a general problem and not related to pluma.

lukefromdc:
In a quick test, I tried rolling GTK3 all the way back to GTK 3.14 just long enough to run gtk3-widget-factory, and still bold or italic default fonts were not applied. Also they were not applied in GTK2's twf.  Changing it in mate-control-center (with GTK 3.23.2) does change the value for "font-name" in org>mate>desktop>interface, but on my box neither GTK3 nor GTK2 nor QT apps apply it with the exceptions of Gparted(applies all) and legacy Firefox ESR(torbrowser) which applies bold but not italic fonts.

raveit65:
Btw, the gsettings key is for mate-desktop package, but where is the code for it?
I like to understand how this font code is working?

lukefromdc:
In mate-desktop there are very few references to fonts outside the "interface" schema, but these two files relate to  gsettings values in general:
libmate-desktop/mate-dconf.c
libmate-desktop/mate-gsettings.c

raveit65:
> ......For me this doesn't even show up in Firefox

This work only with gtk2 version of firefox, eg. in rhel7.

raveit65:
I agree with adding this to pluma as you're maintaining pluma since several years and i respect your work.
But for other MATE applications we need to understand why old font functions doesn't work any more with gtk3 applications, and we need to find a general solution.

lukefromdc:
I am not blocking this, we can revisit it later when a general solution is found. Meanwhile this may benefit someone using Pluma in any environment where bold and italic default font selection actually works.

Issue #301, opened 2018-03-23
pluma crash when trying to open file with BOM at begining


hkalina:
#### Expected behaviour
File with BOM should be opened correctly.

#### Actual behaviour
pluma crash:
Program terminated with signal SIGABRT, Aborted.
#0  _nl_load_domain (domain_file=<optimized out>, domainbinding=<optimized out>) at loadmsgcat.c:1055
1055			    domain->malloced = mem;
[Current thread is 1 (LWP 16478)]
(see downstream report for full stacktrace)

#### Steps to reproduce the behaviour
Try to open attached file test.php using pluma: [test.php.zip](https://github.com/mate-desktop/pluma/files/1841426/test.php.zip)
(note: the problem occure only when file has *.php extension, it is opened correctly when I rename it to *.txt - maybe it is related to syntax highlight)

#### MATE general version
mate-desktop.x86_64   1.19.3-1.fc27

#### Package version
pluma.x86_64    1.19.3-1.fc27

#### Linux Distribution
Fedora 27

#### Link to downstream report of your Distribution
https://bugzilla.redhat.com/show_bug.cgi?id=1558487 (retrace included)

sc0w:
The crash is caused in `pluma-document.c` by the line `gtk_source_buffer_set_language`

rcaridade145:
> the problem occure only when file has *.php extension

Searching a bit on google it seems the recommended fix is to either remove it or ignore it when it is php

[https://stackoverflow.com/questions/3255993/how-do-i-remove-%C3%AF-from-the-beginning-of-a-file](url)

sc0w:
@monsta I don't know.

In https://github.com/mate-desktop/pluma/pull/357 I disabled `gtk_source_buffer_set_language` if the file is `php` with `bom`.

monsta:
I see, it's a workaround, and the actual problem is probably upstream. If we'd find the upstream report, we could mention it in the comments.

PR #357, opened 2018-08-19
pluma-document: Fix: don't crash using files with 'bom'

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/301

### DIFF
--- a/pluma/pluma-document.c
+++ b/pluma/pluma-document.c
@@ -650,12 +650,58 @@ pluma_document_class_init (PlumaDocumentClass *klass)
 	g_type_class_add_private (object_class, sizeof(PlumaDocumentPrivate));
 }
 
+static gboolean
+file_with_bom (GFile *file)
+{
+	FILE    *testfile;
+	gchar    c;
+	int      i;
+	gchar   *bom;
+	gchar   *file_path;
+	gboolean has_bom;
+
+	file_path = g_file_get_path (file);
+
+	testfile = fopen (file_path, "r");
+
+	g_free (file_path);
+
+	if (testfile == NULL)
+	{
+		perror ("fopen");
+		return FALSE;
+	}
+
+	bom = "";
+
+	for (i = 0; i < 3; i++)
+	{
+		c = fgetc (testfile);
+
+		if (c == EOF)
+			break;
+		else
+			bom = g_strdup_printf ("%s%c", bom, c);
+	}
+
+	fclose (testfile);
+
+	if (g_strcmp0 (bom, "\357\273\277") == 0)
+		has_bom = TRUE;
+	else
+		has_bom = FALSE;
+
+	g_free (bom);
+	return has_bom;
+}
+
 static void
 set_language (PlumaDocument     *doc, 
               GtkSourceLanguage *lang,
               gboolean           set_by_user)
 {
 	GtkSourceLanguage *old_lang;
+	const gchar       *bom_langs;
 
 	pluma_debug (DEBUG_DOCUMENT);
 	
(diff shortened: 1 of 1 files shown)


raveit65:
review swap?

lukefromdc:
I have no idea what that is or how to test it. A test file would allow me to test and review this

sc0w:
@lukefromdc the mentioned issue in my first comment has a test file: https://github.com/mate-desktop/pluma/files/1841426/test.php.zip

lukefromdc:
Error was same as with master and with gedit, I could not open the test file in any text editor except nano (in terminal).

sc0w:
@lukefromdc please try again.

> I could not open the test file in any text editor except nano (in terminal).

It works in geany

lukefromdc:
I teied two builfs, once with the changed file dropped into a pluma build directory, once with a new build from the git branch. Has there been a forced update to it? I am on the road now. If not, maybe GTK 3.23 is the difference? That I can test when I get home

sc0w:
@lukefromdc I did little changes, I confirm it didn't work before, now it must work.

`== 0` is needed with `g_strcmp0` , this was my fault.

monsta:
Ok, we react to malformed `php` with BOM at the start... what about malformed `xml` with BOM? :slightly_smiling_face: 

[bad.xml.zip](https://github.com/mate-desktop/pluma/files/2306529/bad.xml.zip)

```

**Comment by sc0w** [comment]
@monsta done!

**Comment by monsta** [comment]
Well, the point is that the problem is in `GtkSourceView`. We can try working it around for one file format, two file formats... but the actual problem is their abort.

```C
g_assert (g_slist_length (ce->priv->invalid) <= 1);
```

This is the source of the mentioned error message:
```
GtkSourceView:ERROR:gtksourcecontextengine.c:5543:update_syntax: assertion failed: (g_slist_length (ce->priv->invalid) <= 1)
```

Basically, it means: "oh, something is wrong in our **internal** structure, so let's abort the whole app that uses our library."

lukefromdc:
A possible way to catch that would be to use a conditional check matching the assert `(g_slist_length (ce->priv->invalid) <= 1)` and whenever this returns FALSE don't run `gtk_source_buffer_set_language` rather than special casing file types

monsta:
I don't think we have access to that, `ce->priv` is internal to the library.

lukefromdc:
Is there anything we are sending to gtk_source_buffer_set_language other than the file type that could be used as a marker for this to detect this error?

sc0w:
For me this is a critical bug, we can loss important files without changes, if pluma was already opened before open the 'killer' files.

@lukefromdc 

> Is there anything we are sending to gtk_source_buffer_set_language other than the file type that could be used as a marker for this to detect this error?

**UTF-8**

I mean:

`pluma test.php --encoding UTF-8` <- it crashes


`pluma test.php --encoding ISO-8859-1` <- it doesn't crash

you can see the list of encodings with:

`pluma --list-encodings`

lukefromdc:
What would we lose if we removed gtk_source_buffer_set_language entirely? Dictionary support? Or could this be made manually selectable or otherwise made to work

sc0w:
@lukefromdc we loss the highlighting, with this PR `php` and `xml` files with `bom` can be opened, but without the highlighting.

sc0w:
...and `php xml` files without `bom` works the highlighting with this PR.

lukefromdc:
OK,is there any way we could reimplement this locally? The highlighting is a crucial feature, and GTK has made some ugly choices. If not, we will just have to find and exclude all the killer file types as they come up, and leave it at that.

sc0w:
oops, there are more formats affected, this is the complete list:

```
ASP
django template
docboox
HTML
MXML
Mallard
Markdown
MediaWiki
PHP
Tera template
XML
XSLT
```

sc0w:
I added all the affected languages in latest changes

lukefromdc:
At the moment I can open XML files fine, I assume this change applies only to files starting with ﻿<? ?

lukefromdc:
Than this looks good to go

Issue #356, opened 2018-08-17
unsaved changes dialog often requires selecting "Close without Saving" twice


stiunnif:
#### Expected behaviour
when "Close without Saving" button is selected, the dialog and pluma window should immediately close.

#### Actual behaviour
often, when the "Close without Saving" button is selected, the unsaved changes dialog comes up a second time (or perhaps the original dialog is modified in place)

note: the "second dialog" (if it is in fact a second dialog) is about a line shorter than the first one, if that's some clue. also, in instances where the second instance of the dialog isn't persistently shown, a keen eye may notice that the shorter second dialog flashes quickly before pluma closes out.

also, the second dialog has "Close without Saving" highlighted, rather than the original default of "Save", which may be another clue (e.g., is this a second dialog vs is it the same dialog but modified)

i didn't used to encounter this issue in pluma, but it's been going on for quite a while now (several months, or perhaps even a year or more)

#### Steps to reproduce the behaviour
note: the second dialog does not consistently appear persistently (although it does seem to consistently get shown momentarily in instances where it doesn't persist, as described above) ...

prep: 

- create file named "docOne" with content: "[original docOne text]"
- create file named "docTwo" with content: "[original docTwo text]"
- save these files and close pluma

1. open "docOne"
2. type something in "docOne" without saving
3. open "docTwo"
4. type something in "docTwo" without saving
5. close pluma - select "Close without Saving" - watch carefully to see second (shorter) instance of unsaved changes dialog flash before the pluma window closes (or, the second instance may persist, requiring you to select "Close without Saving" a second time)

screen capture showing first unsaved changes dialog:
![first-dialog](https://user-images.githubusercontent.com/42481165/44291027-c074ea00-a249-11e8-8950-8089414d64f2.png)

screen capture showing second unsaved changes dialog:
![second-dialog](https://user-images.githubusercontent.com/42481165/44291032-c965bb80-a249-11e8-8838-932f38a9359c.png)

#### MATE general version
1.20.2

#### Package version
1.20.2-2

#### Linux Distribution
Fedora 28

lukefromdc:
With Pluma 1.21.1 from git master running on Debian Unstable I was unable to duplicate this, wonder what is different?

sc0w:
I can reproduce the issue (old issue for me) in pluma 1.21.1

lukefromdc:
Oh THAT's interesting. No theme should be able to have any effect on this kind of bug. 

[...]

PR #358, opened 2018-08-20
close-confirmation-dialog: Fix: don't resize the dialog unexpectedly

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/356


lukefromdc:
Memory management should be fine with this. Correct me if I am wrong but it looks like the two new variables are called on the stack (not declared with *) and thus do not need to be freed

sc0w:
I think int variables don't need no be freed

lukefromdc:
If my understanding is correct, an integer declared as int does not have to be freed, declared as *int is dynamically allocated and if not freed leaks when the function is closed.

PR #359, opened 2018-08-21
make the 'close confirmation dialog' bigger with multiple files

sc0w:
how to test: just open multiple files to force the dialog and see the size:

screenshot at 2018-08-21 18-59-34

Now resizable if it has vertical scrollbar, and the max height of the dialog is the 50% of the height of monitor (I can increase it if you like more).

### DIFF
--- a/pluma/dialogs/pluma-close-confirmation-dialog.c
+++ b/pluma/dialogs/pluma-close-confirmation-dialog.c
@@ -674,17 +674,22 @@ static void
 build_multiple_docs_dialog (PlumaCloseConfirmationDialog *dlg)
 {
 	PlumaCloseConfirmationDialogPrivate *priv;
-	GtkWidget *hbox;
-	GtkWidget *image;
-	GtkWidget *vbox;
-	GtkWidget *primary_label;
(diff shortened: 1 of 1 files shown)


raveit65:
Personal i would prefer a modern widescreen format of the dialog.
But your decision.......
pluma

sc0w:
@raveit65 what means "a modern widescreen format of the dialog" ?

more/less width?

more/less height?

percentages over the monitor dimensions?

raveit65:
@sc0w 
I mean with your new default size the dialog looks like a quarter.
Most computer monitors are using 16:9 proportion.
So, less height looks better for me, see my screenshot.

sc0w:
I changed 50% to 40% of height monitor



You can test other values here, changing 40 to other:
https://github.com/mate-desktop/pluma/pull/359/files#diff-3f76460d36df6317350829111799002aR793

@raveit65 
Now it looks better in your monitor?

raveit65:
```
max_height = mon_geo.height * 40 / 100;
```
is much better :)

Issue #179, opened 2016-04-06
Don't wipe PRIMARY selection when unselecting text


haarp:
One of the greatest idiocies of Gnome is the fact that the PRIMARY selection (middle mouse button buffer) gets wiped when text is deselected.
- Select text -> primary selection is filled
- Deselect text (**without** selecting new text) -> primary selection is emptied

Why?!

Pluma should fix this.
##

haarp:
Turns out this is a GTK problem.

https://bugzilla.gnome.org/show_bug.cgi?id=584236

haarp:
Awesome, thanks!

PR #360, opened 2018-08-23
pluma-view: Fix: Don't loss the PRIMARY clipboard

sc0w:
Fixes https://github.com/mate-desktop/pluma/issues/179

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -2137,6 +2137,18 @@ show_line_numbers_menu (GtkWidget      *view,
 static gboolean
 pluma_view_button_press_event (GtkWidget *widget, GdkEventButton *event)
 {
+	static gchar  *primtxt = "";
+
+	gchar *txt_clip = gtk_clipboard_wait_for_text (gtk_clipboard_get (GDK_SELECTION_PRIMARY));
+
+	if (txt_clip)
+	{
+		primtxt = g_strdup (txt_clip);
+		g_free (txt_clip);
+	}
+	else
+		gtk_clipboard_set_text (gtk_clipboard_get (GDK_SELECTION_PRIMARY), primtxt, strlen (primtxt));
+
 	if ((event->type == GDK_BUTTON_PRESS) && 
 	    (event->window == gtk_text_view_get_window (GTK_TEXT_VIEW (widget),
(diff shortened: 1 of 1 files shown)


lukefromdc:
With my fingers, a way to empty this buffer, select NULL, or simply disable middle click paste is absolutely necessary. Worst case I can patch Pluma locally to remove it, but any way of emptying the buffer when deselecting text does not do it will be good enough for me. I had to disable almost all the middlemouse click features in Firefox for similar reasons on my boxes.

Maybe a gsettings option if there is not already a way to empty the buffer?

lukefromdc:
GNOME does have the ability (in gnome-tweak-tool's "keyboard and mouse" to enable or disable middle click paste. Maybe we could add that to M-C-C?

sc0w:
@lukefromdc  I will tray to do something in other PR

lukefromdc:
I am now looking at 
https://github.com/GNOME/gnome-settings-daemon/commit/aad836ebcbf2a20f72188ced1cc8fe33da6d39ac#diff-98eb5b265d77a75d070ec322831ebe58
 from gnome-settings-daemon as a possible way to fix this issue for myself and anyone else needing to be able to turn middlemouse paste on and off. Default should be ON as that is the normal Xorg default(this is not available in Wayland at all). When I have it ready I will submit a PR and test it with Pluma and this PR applied

lukefromdc:
Also turn out adding 
gtk-enable-primary-paste=false
to ~/.local/gtk-3.0/settings.ini will turn this off, I just confirmed this to work by testing. Now let's see if that GNOME commit works and lets dconf-editor do it, if so we can then look at a GUI checkbox for m-c-c

lukefromdc:
Meanwhile I withdraw all objections to this PR as the problem has an existing solution

lukefromdc:
Please test
https://github.com/mate-desktop/mate-desktop/pull/332
and
https://github.com/mate-desktop/mate-settings-daemon/pull/234
together, they allow enabling or disabling middlemouse paste from dconf-editor. Default is "enabled" to avoid any change in expected behavior with a default install.

lukefromdc:
Also please test 
https://github.com/mate-desktop/mate-control-center/pull/369
which adds a checkbox in mate-control-center>mouse to enable(default) or disable middlemouse paste

PR #361, opened 2018-08-24
pluma-document: fix possible memory leak

sc0w:
In the commit https://github.com/mate-desktop/pluma/commit/4a74dc65ffb89e0ef391cdf1f0d4c07cb1d20889

The line `bom = g_strdup_printf ("%s%c", bom, c);` is called three times, but freed one time

https://github.com/mate-desktop/pluma/commit/4a74dc65ffb89e0ef391cdf1f0d4c07cb1d20889#diff-f46f2cae08ed2899ec953fd47b56976eR684

so, I did little changes to make sure there is no memory leaks.

### DIFF
--- a/pluma/pluma-document.c
+++ b/pluma/pluma-document.c
@@ -656,9 +656,8 @@ file_with_bom (GFile *file)
 	FILE    *testfile;
 	gchar    c;
 	int      i;
-	gchar   *bom;
+	gchar    bom[3];
 	gchar   *file_path;
-	gboolean has_bom;
 
 	file_path = g_file_get_path (file);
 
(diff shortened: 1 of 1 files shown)


sc0w:
I will squash this PR and https://github.com/mate-desktop/pluma/pull/357 in one commit to the 1.20 branch

lukefromdc:
Wouldn't that require a forced update to master? After that, all local repos would have to be repaired by resetting to a prior commit (before the first one changed in the forced update) and pulling again to restore. Simple enough for me to do that, what about others here?

sc0w:
@lukefromdc no, I never did that to official branches, to the 1.20 branch I just will push one new commit.

sc0w:
seems my english is not clear: I never will do that!

lukefromdc:
Thanks, I was not thinking of a branch yet to receive the earlier commit but rather of master

PR #368, opened 2018-09-23
Use actual pluma screenshot

rbuj:


raveit65:
I just want to mention that Menta (green) is our default theme. But replacing all images in all languages is a very huge job and not needed for this reason.
So, personal i am fine with using TraditionalOK theme for replacing images.

raveit65:
And cherry-picked to 1.20.
Thank you.

PR #370, opened 2018-09-23
Use actual pluma screenshot for Catalan

rbuj:


lukefromdc:
I don't have that locale installed (all unused locales purged), but this does not create any issues with the help for Pluma coming up and builds fine

Issue #124, opened 2015-07-07
0x0A added to EOF


ge79so:
Pluma adds a linefeed (0x0A) to the end of files when saving. 

This causes problems with string comparison in programming (eg PHP's file_get_contents() function) since a one-line file created with pluma does not consist of "line<eof>" as expected, but "line<newline><eof>"

Can you fix this?

monsta:
I can't reproduce this, except for the case where sort plugin was used on the file contents (https://github.com/mate-desktop/pluma/issues/153).

ge79so:
The problem occurs with Pluma 1.12.2 on Ubuntu 16.04.2 - please see the attached screenshot - the sort plugin is not installed (/ticked)

![selection_002](https://cloud.githubusercontent.com/assets/10240950/23509770/d583ce06-ff4e-11e6-8cea-a114d356f8a8.jpg)

monsta:
Ah, I see it now. It does add 0x0A but it's not displayed as the newline for some reason. Only using a hex viewer would show it.

Adding a newline at the end of file results in two 0x0A symbols at the end. Weird.

monsta:
Ok, so this behavior was introduced long ago in gedit:

https://git.gnome.org/browse/gedit/commit/?id=127350ec8c1ffb31340e183ace144ee3dad40c06
https://bugzilla.gnome.org/95676
https://bugzilla.gnome.org/324942

maravento:
Same problem (ubuntu Mate 18.04). trailing-newline to the end of file. How to fix it?
In "gedit" it can be eliminated, with run dconf-editor and /org/gnome/gedit/preferences/editor
change "ensure-trailing-newline" to false
But "pluma" does not have this option to disable it

lukefromdc:
That sounds like there is a gedit commit that can be applied to Pluma to fix this

maravento:
@lukefromdc ... And what is the solution?

lukefromdc:
I do not know, I haver not had time to go trolling through gedit commits looking for it. My last couple weeks have been very busy

maravento:
someone else posted the problem: https://askubuntu.com/questions/1048360/saving-files-without-a-trailing-newline-in-pluma-text-editor

gamax92:
The relevant commit for gedit is GNOME/gedit@fe62ab8175747298a5315aaeed53ef0c56e60817

alexander255:
[...]

PR #388, opened 2018-12-18
Don't add and trim trailing newlines from documents

ntninja:
Fixes issue #124.

### DIFF
--- a/pluma/pluma-document-input-stream.c
+++ b/pluma/pluma-document-input-stream.c
@@ -45,7 +45,6 @@ struct _PlumaDocumentInputStreamPrivate
 
 	PlumaDocumentNewlineType newline_type;
 
-	guint newline_added : 1;
 	guint is_initialized : 1;
 };
 
@@ -437,27 +436,6 @@ pluma_document_input_stream_read (GInputStream  *stream,
 					  &iter,
 					  dstream->priv->pos);
 
-	if (gtk_text_iter_is_end (&iter) &&
-	    !gtk_text_iter_is_start (&iter))
-	{
-		gssize newline_size;
-
-		newline_size = get_new_line_size (dstream);
-
-		if (space_left >= newline_size &&
-		    !dstream->priv->newline_added)
-		{
-			const gchar *newline;
-
-			newline = get_new_line (dstream);
-
-			memcpy ((void *) ((gsize) buffer + read), newline, newline_size);
-
-			read += newline_size;
-			dstream->priv->newline_added = TRUE;
-		}
-	}
-
 	return read;
 }
 
@@ -467,8 +445,6 @@ pluma_document_input_stream_close (GInputStream  *stream,
 				   GError       **error)
 {
 	PlumaDocumentInputStream *dstream = PLUMA_DOCUMENT_INPUT_STREAM (stream);
-
-	dstream->priv->newline_added = FALSE;
 
 	if (dstream->priv->is_initialized)
 	{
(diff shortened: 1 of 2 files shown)


lukefromdc:
How do I test this?

alexander255:
Good point! :slightly_smiling_face: 

Here's a testing scenario:

1. Start `pluma` without this patch applied
2. Open/create a text file, write a single line of text with no line breaks and save the document
3. Close `pluma`
4. Open the text file in a hex editor (such as `ghex`)
5. Observe that additionally to the content you wrote a newline character (0x10) was appended to the contents
6. Close the hex editor again
7. Open the same text file in patched `pluma` with these changes applied
8. Observe that the final newline is now visible
9. Remove the final newline and save again
10. Close `pluma` again
11. Open the text file in the hex editor again
12. Observe that now only the content was written without any extra characters

lukefromdc:
With pluma from master I saved this file (with no newline in it) 
`1234566789010`
and got this hexdump from it:

```
0000000 3231 3433 3635 3736 3938 3130 0a30     
000000e

```

alexander255:
The output is correct, but you apparently set the Hex editor to 16-bit little endian :slightly_smiling_face: 
See how it groups each set of two characters (16-bit) and reverses their order (little endian)? Try setting it to 8-bit and it should be more obvious what is happening!

alexander255:
Also it appears I wrote some nonsense into my scenario: The newline character is of course `0x0A` (=10d), not 0x10 (= 16d). Sorry bout that!

lukefromdc:
Also note that for testing your test files need to be created with the version of pluma being tested

alexander255:
@lukefromdc: Thanks for trying to review this!

A note regarding your last comment: You can also create the testing files without switching pluma version (`printf 'bla\n' > test-file` vs `printf 'bla' > test-file`), but then you cannot test the whole load+edit+save cycle making it rather useless.

lukefromdc:
@monsta @raveit65 , are either of you able to properly evaluate this?

raveit65:
Ok that works, with PR the newline is visible in pluma and can be removed.
But why this is a problem?
And maybe we can follow gedit upstream and remove the whole file, as all is now GtkSourceView?
See https://gitlab.gnome.org/GNOME/gedit/commit/f7607b420e52836b5c177d3835185bcf562fb7ac
@monsta @sc0w 
Thoughts?

sc0w:
I noticed vi / vim works like pluma without this PR, and it shows a warning opening files without new line at the end:



why?

we need a gsettings key to enable the new behavior of this PR?

sc0w:
and github shows warning adding new files with no new line at the end:

raveit65:
> and github shows warning adding new files with no new line at the end:

This isn't good

sc0w:
I agree with the PR if it adds a gsettings key

gedit has this key:

sc0w:
it was added in gedit with the commits:

https://gitlab.gnome.org/GNOME/gedit/commit/fe62ab8175747298a5315aaeed53ef0c56e60817

https://gitlab.gnome.org/GNOME/gedit/commit/42a8884a2b9b58bf5216c608f8266fde2af280b6

sc0w:
`git diff` also shows warning with no new line at the end:

raveit65:
Any news or progress?
Or should we close PR?

raveit65:
Closing, ask for reopening if you want to work again on it.

Issue #391, opened 2019-01-18
Bug in size of snippets Window


andrewfowlie:
#### Expected behaviour

When I select the Tools->Manage Snippets... window, the first time, the size of the window is alright. I expect it to be the size I left it with when I close it then again select Tools->Manage Snippets...

#### Actual behaviour

Any subsequent times I do it the Manage Snippets window is too small to see the contents and I have to manually expand it.

#### Steps to reproduce the behaviour

Tools->Manage Snippets... 
Note the size. Close the window
Tools->Manage Snippets... 
Note the tiny size

#### MATE general version

Distributor ID:	Ubuntu
Description:	Ubuntu 18.10
Release:	18.10
Codename:	cosmic

#### Package version

pluma - Version 1.20.2

#### Linux Distribution

Linux desktop 4.18.0-10-generic #11-Ubuntu SMP Thu Oct 11 15:13:55 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

# Fix

The problem is that the lines to save the current size of the window

https://github.com/mate-desktop/pluma/blob/94b306143618cfbd9a0c8908ea7d47f3a86ef6af/plugins/snippets/snippets/Manager.py#L602-L603

get called after the relevant object gets destroyed

https://github.com/mate-desktop/pluma/blob/94b306143618cfbd9a0c8908ea7d47f3a86ef6af/plugins/snippets/snippets/Manager.py#L616

as a result, a default size of (1, 1) from `get_allocate` on a destroyed object gets returned and used the default. I will push a make a PR with a fix that gets the size before the object gets destroyed.

raveit65:
Confirmed with fedora 28.

tamplan:
This issue should be closed by now.

PR #392, opened 2019-01-18
addresses #391 - gets snippet window size before destroying it

andrewfowlie:
addresses #391.

You see I moved the code to get the window size to before the point where it's destroyed destroyed.

I could have done it the other way - move the destroyed bit to after the window size operations. But moving where something is destroyed seemed more dangerous than just moving something about the default window size, especially as I don't know the structure of the code.

### DIFF
--- a/plugins/snippets/snippets/Manager.py
+++ b/plugins/snippets/snippets/Manager.py
@@ -599,16 +599,17 @@ class Manager:
                 if self.snippets_doc:
                         self.snippets_doc.stop()
                 
-                alloc = dlg.get_allocation()
-                self.default_size = [alloc.width, alloc.height]
                 self.manager = None
-
                 self.unref_languages()        
                 self.snippet = None        
                 self.model = None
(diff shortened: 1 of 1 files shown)


raveit65:
Easy fix, thanks for pointing that out.

raveit65:
Can you please avoid wirting a link in commit header next time?
This belongs in commit body.
Something like this looks really better when you browsing through git history.
```
From 2b00f658a9162af62a1318ffd4dcc21563afe198 Mon Sep 17 00:00:00 2001
From: Andrew Fowlie <andrew.j.fowlie@qq.com>
Date: Fri, 18 Jan 2019 16:30:29 +0800
Subject: [PATCH] gets snippet window size before destroying it

addresses #391
```

andrewfowlie:
@raveit65  sure no worries.

andrewfowlie:
@raveit65  sure no worries.

PR #397, opened 2019-02-06
Upgrade the manual to docbook 5.0

rbuj:
To upgrade the manual:
  sudo dnf install -y docbook5-schemas
  xsltproc --output index-new.docbook /usr/share/xml/docbook5/stylesheet/upgrade/db4-upgrade.xsl index.docbook
  xsltproc --output legal-new.xml /usr/share/xml/docbook5/stylesheet/upgrade/db4-upgrade.xsl legal.xml

To validate the manual:
  xmllint --noout --relaxng /usr/share/xml/docbook5/schema/rng/5.0/docbook.rng help/C/index.docbook
  jing /usr/share/xml/docbook5/schema/rng/5.0/docbook.rng help/C/index.docbook
  yelp-check validate help/C/index.docbook

To view the manual:
  yelp file:///full_path/index.docbook

Note: docbook5-schemas should be installed in your system in order to view the manual (DEPENDENCY)

### DIFF
--- a/help/C/index.docbook
+++ b/help/C/index.docbook
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<?xml-model href="http://docbook.org/xml/5.0/rng/dbits.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
 <!--
       (Do not remove this comment block.)
   Maintained by the MATE Documentation Project
@@ -18,21 +19,21 @@
     <info>
         <title>Pluma Manual</title>
 
-        <copyright>
+        <copyright its:translate="no">
             <year>2015</year>
             <holder>MATE Documentation Project</holder> 
         </copyright>
-        <copyright>
+        <copyright its:translate="no">
             <year>2007</year>
             <holder>GNOME Documentation Project</holder> 
         </copyright>
-        <copyright>
+        <copyright its:translate="no">
             <year>2002</year>
             <year>2003</year>
             <year>2004</year>
             <holder>Sun Microsystems</holder> 
         </copyright>
-        <copyright>
+        <copyright its:translate="no">
             <year>2000</year>
             <holder>Eric Baudais</holder> 
         </copyright>
@@ -46,11 +47,11 @@
 
         <xi:include href="legal.xml"/>
 
-        <authorgroup>
+        <authorgroup its:translate="no">
             <author>
-                <orgname>MATE Documentation Project</orgname>
+                <orgname its:translate="yes">MATE Documentation Project</orgname>
                 <affiliation>
-                    <orgname>MATE Desktop</orgname>
+                    <orgname its:translate="yes">MATE Desktop</orgname>
                 </affiliation>
             </author>
             <author>
@@ -60,9 +61,9 @@
                 </personname>
             </author>
             <author>
-                <orgname>GNOME Documentation Project</orgname>
+                <orgname its:translate="yes">GNOME Documentation Project</orgname>
                 <affiliation>
-                    <orgname>GNOME</orgname>
+                    <orgname its:translate="yes">GNOME</orgname>
                 </affiliation>
             </author>
             <author>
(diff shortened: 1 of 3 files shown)


raveit65:
Sounds like we need a dependency check for docbook5-schemas?
As package maintainer i can add a run time require to fedora rpm, but what happens if someone forget this if installing by hand?
I guess the manual won't work anymore with older or docbook-4.0 ?

rbuj:
I'm sorry. It's only required for yelp-check, as it uses --noent.

https://github.com/GNOME/yelp-tools/blob/2f0118eea1fc0921252efd389bd74da95b949f5c/tools/yelp-check.in#L1067

$ yelp-check validate help/C/index.docbook 
error : Unknown IO error
warning: failed to load external entity "http://docbook.org/xml/5.0/rng/docbook.rng"
Relax-NG parser error : xmlRelaxNGParse: could not load http://docbook.org/xml/5.0/rng/docbook.rng
Relax-NG schema http://docbook.org/xml/5.0/rng/docbook.rng failed to compile

raveit65:
So we can't break distros a few days before feature-freeze for 1.20 release?
Maybe it's better to keep it for next dev cycle (1.23) ?

monsta:
Yes it's quite big... so if we don't require this for 1.22, we can postpone it.

I don't see yelp-check called anywhere during the build. So, do we need the schemas package after all?

rbuj:
> Yes it's quite big... so if we don't require this for 1.22, we can postpone it.
> 
> I don't see yelp-check called anywhere during the build. So, do we need the schemas package after all?

docbook5 schemas need to be installed in order to run make check-help

https://github.com/GNOME/yelp-tools/blob/2f0118eea1fc0921252efd389bd74da95b949f5c/tools/yelp.m4#L144

raveit65:
Only rebased against master.

raveit65:
Pluma help opens fine with this PR but without having `docbook5-schemas` installed.
So when i install docbook5-schemas i need to update index.docbook and legal.xml in an existing installation, right?
What happens if in install `docbook5-schemas` without updating index.docbook and legal.xml?
Isn't it better to force the installation of `docbook5-schemas`with adding it as dependency, before install of index.docbook and legal.xml?

raveit65:
With docbook5-schemas installed.
```
[rave@mother pluma-dist]$ yelp-check validate help/C/index.docbook
help/C/index.docbook:22: element copyright: Relax-NG validity error : Expecting an element holder, got nothing
help/C/index.docbook:22: element copyright: Relax-NG validity error : Invalid attribute translate for element copyright
Relax-NG validity error : Extra element copyright in interleave
help/C/index.docbook:22: element copyright: Relax-NG validity error : Element info failed to validate content
help/C/index.docbook:19: element info: Relax-NG validity error : Element article failed to validate content
help/C/index.docbook fails to validate
```
This works now and give us a new TODO list -_-
And pluma help opens fine without migrating index.docbook and legal.xml to docbook5.
So, it seems it is save to use this update in any case.

But i am not sure if i understand for what it is good for, if it doesn't have any influence of using pluma help.

rbuj:
Docbook V4.x is no longer being actively developed. Docbook V4.5 version was published on [October 20, 2006](https://docbook.org/schemas/4x). Docbook has new stuff, including [ITS 2.0](https://www.w3.org/TR/its20/) for translations. Most tools focus their development to support Docbook 5.

raveit65:
Ok, but....
How can we make sure that `docbook5-schemas` is installed and used?
Otherwise it doesn't matter if we install index.docbook and legal.xml for docbook5, or not ?
Or is this only for developing needed? (yelp-check validate.....)

rbuj:
db5 schema it's only required for development, yelp parses docbook files to html on runtime using xslt.

raveit65:
> db5 schema it's only required for development, yelp parses docbook files to html on runtime using xslt.

Very important info for a review ;)

raveit65:
Damned, no way to cherry-pick it easy to 1.22 because update of .pot files are in the same commit.

yetist:
[Mallard](http://projectmallard.org/index.html) is a markup language to provide user help documents, and [yelp](https://wiki.gnome.org/Apps/Yelp/Mallard) support it.
Willing to try?

raveit65:
We had a discussion 2 years ago about switching to `mallard`.
This means a complete rewrite of the whole user-guide, which is a very huge job for several months with several people.
There was a guy who started with it https://github.com/mate-desktop-legacy-archive/mate-user-guide-mallard
After 2 months he left the team without any word. ;)
I am in doubt that we have the manpower for it.

yetist:
mate-user-guide is really a huge job, let it continue to use docbook. If we plan to use meson, we may have to think about this problem, as far as I know, meson's gnome.yelp only supports mallard.

lukefromdc:
One way to bypass this for meson would be this: move all of the help files from all other packages into mate-user-guide, keep building m-u-g with autotools, but use meson for the rest of the packages

PR #401, opened 2019-02-16
Enable Travis CI support

yetist:
This PR maybe OK, but I found an issue when building pluma on archlinux, I will open a new PR to fix that.


yetist:
New PR #402  created to fix build fail on archlinux.

PR #402, opened 2019-02-16
Fix build fail on Archlinux

yetist:
The enchant on Archlinux is 2.2.3-2, that include enchant-2.pc.
This commit first check enchant-2, then check enchant.

Here can see the file list of enchant on archlinux:
https://www.archlinux.org/packages/extra/x86_64/enchant/

```
usr/lib/libenchant-2.so
usr/lib/libenchant-2.so.2
usr/lib/libenchant-2.so.2.2.3
usr/lib/pkgconfig/
usr/lib/pkgconfig/enchant-2.pc
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,7 +98,7 @@ dnl ================================================================
 dnl spell plugins checks: enchant and iso-codes
 dnl ================================================================
 
-ENCHANT_REQUIRED=1.2.0
+ENCHANT_REQUIRED=1.6.0
 ISO_CODES_REQUIRED=0.35
 
 AC_ARG_ENABLE([spell],
@@ -107,15 +107,19 @@ AC_ARG_ENABLE([spell],
 	[enable_enchant=yes])
 
 if test "x$enable_enchant" = "xyes" ; then
-
-	PKG_CHECK_MODULES(ENCHANT, enchant-2 >= $ENCHANT_REQUIRED, \
-			  have_enchant_2=yes, have_enchant_2=no)
-
-	if test "x$have_enchant_2" = "xno"; then
-                PKG_CHECK_MODULES(ENCHANT, enchant >= $ENCHANT_REQUIRED, \
-                                  have_enchant=yes, have_enchant=no)
+        ENCHANT_MODNAME=enchant
+        PKG_CHECK_EXISTS([enchant-2 >= $ENCHANT_REQUIRED], [have_enchant_2=yes], [have_enchant_2=no])
+        if test "x$have_enchant_2" = "xyes"; then
+                ENCHANT_MODNAME=enchant-2
         fi
-	if test "x$have_enchant_2" = "xyes" || test "x$have_enchant" = "xyes"; then
+
+        AC_MSG_CHECKING([the version of enchant])
+        ENCHANT_VERSION=`pkg-config --modversion $ENCHANT_MODNAME`
+        AC_MSG_RESULT([$ENCHANT_VERSION])
+
+        PKG_CHECK_MODULES(ENCHANT, [$ENCHANT_MODNAME >= $ENCHANT_REQUIRED], \
+                          have_enchant=yes, have_enchant=no)
+	if test "x$have_enchant" = "xyes"; then
 
 		PKG_CHECK_EXISTS([iso-codes >= $ISO_CODES_REQUIRED],
 				 [have_iso_codes=yes],[have_iso_codes=no])
--- a/configure.ac
+++ b/configure.ac
@@ -108,10 +108,14 @@ AC_ARG_ENABLE([spell],
 
 if test "x$enable_enchant" = "xyes" ; then
 
-	PKG_CHECK_MODULES(ENCHANT, enchant >= $ENCHANT_REQUIRED, \
-			  have_enchant=yes, have_enchant=no)
+	PKG_CHECK_MODULES(ENCHANT, enchant-2 >= $ENCHANT_REQUIRED, \
+			  have_enchant_2=yes, have_enchant_2=no)
 
-	if test "x$have_enchant" = "xyes"; then
+	if test "x$have_enchant_2" = "xno"; then
+                PKG_CHECK_MODULES(ENCHANT, enchant >= $ENCHANT_REQUIRED, \
+                                  have_enchant=yes, have_enchant=no)
+        fi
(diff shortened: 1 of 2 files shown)


raveit65:
Interesting, fedora chips also enchant2
```
[rave@mother ~]$ dnf list enchant2\*
Last metadata expiration check: 0:06:20 ago on Sat Feb 16 17:08:26 2019.
Installed Packages
enchant2.x86_64                          2.2.3-6.fc30                   @rawhide
Available Packages
enchant2.i686                            2.2.3-6.fc30                   rawhide 
enchant2-aspell.x86_64                   2.2.3-6.fc30                   rawhide 
enchant2-devel.i686                      2.2.3-6.fc30                   rawhide 
enchant2-devel.x86_64                    2.2.3-6.fc30                   rawhide 
enchant2-voikko.x86_64                   2.2.3-6.fc30                   rawhide
```
Still wondering that eom can simply build against another library?
Btw. How did archlinux maintainer build eom all the time?
I mean, is this tested at archlinux?

Not sure if such an update needs more testing before 1.22 release and should better be delayed.
@monsta @lukefromdc @sc0w 
Thoughts?

monsta:
Wow, I had no idea. Indeed, how do you build it then without this PR?

Debian and Ubuntu still have 1.6.0. There is 2.1.2 in the [experimental](https://packages.debian.org/experimental/libenchant-dev) repo of Debian, but the filelist there still shows me `/usr/lib/x86_64-linux-gnu/pkgconfig/enchant.pc`.

monsta:
Need to check if more distros would need this, to decide if we add this to 1.22 or tell these distros to use a patch...

raveit65:
If it is needed for archlinux i think we should add it.
But i would like to see that someone can confirm that pluma runs well in archlinux.
Well, i will test it local with a fedora build.
Let see if i can damage my system :P

yetist:
I just resolved the build issue for archlinux, and then I will test pluma on archlinux.

Before 1.22 release, we should check all other repos that use enchant, but I didnt clone all the repos of mate-desktop, if who can tell me which used anchant, will be helpful.

sc0w:
It seems only `pluma` needs `libenchant` in the build

lukefromdc:
This should build on all distros before release UNLESS there is a problem in the distro itself blocking build. A broken package in a distro should not block our own release, especially if that distro is not itself about to go to a release version with the offending package unfixed. Yes, we will need to advise a patch in such cases, and lets not forget we can do a point release at any time

yetist:
I am running pluma on archlinux for test, when I start to enable "Spell Checker" plugin in the "Pluma Preferences", I get the following warning:
```
$ pluma

(pluma:1139): Gtk-WARNING **: 14:03:42.551: Theme parsing error: <data>:1:17: Expected a string.

(pluma:1139): libpeas-WARNING **: 14:03:51.556: Failed to load module 'spell': /usr/lib/pluma/plugins/libspell.so: undefined symbol: enchant_dict_add_to_pwl

(pluma:1139): libpeas-WARNING **: 14:03:51.556: Error loading plugin 'spell'
```
https://github.com/winlibs/enchant/blob/master/src/enchant.c#L1040 show  `enchant_dict_add_to_pwl` is deprecated,  should use `enchant_dict_add` instead.
I updated this PR, but need more test on other distros.

monsta:
So they removed that deprecated function in 2.0.0...
https://github.com/AbiWord/enchant/commit/744a1a19c1d0a5c3e1359fa994cf60a4d98ee12e#diff-9eef52f244d01627950b4b2debb7ad92

But even in 1.6.0 that function just called `enchant_dict_add` with the same arguments, so it should be ok to call it directly from our code.

monsta:
I think we can bump the required version of enchant to 1.6.0 just in case.

raveit65:
Spell-checker plugin builds and runs fine with both version of `enchant`.
I don't get any related build or run time warnings.
So this is +1
I would give +++1 if we could see in configure build log which version of  enchant is used :)
```
checking for xgettext... (cached) /usr/bin/xgettext
checking for ENCHANT... yes
checking whether iso-codes has iso-639 and iso-3166 domains... yes
```

yetist:
- Indent code.
- Update the required version of enchant to 1.6.0.
- Show the enchant version on configure output like this:
```
checking for xgettext... (cached) /usr/bin/xgettext
checking the version of enchant... 2.2.3
checking for ENCHANT... yes
checking whether iso-codes has iso-639 and iso-3166 domains... yes
```

eli-schwartz:
Hmm :) https://bugs.archlinux.org/task/61618

eli-schwartz:
BTW apparently the archlinux repository build succeeds just fine:

```
  CC       pluma-spell-checker.lo
pluma-spell-checker.c: In function ‘pluma_spell_checker_add_word_to_personal’:
pluma-spell-checker.c:436:2: warning: implicit declaration of function ‘enchant_dict_add_to_pwl’; did you mean ‘enchant_dict_add_to_session’? [-Wimplicit-function-declaration]
  enchant_dict_add_to_pwl (spell->dict, word, len);
  ^~~~~~~~~~~~~~~~~~~~~~~
  enchant_dict_add_to_session
```

No errors there.

There are hundreds of lines of deprecated gtk functions, so it's not feasible to notice the missing symbol just by checking to see if the build generated warnings.

sc0w:
@eli-schwartz this PR as a patch fixes https://bugs.archlinux.org/task/61618 ?

yetist:
Yes, with 87df43f876abd0f74ceecfb6ec3812edac30381f, pluma can build and can run on archlinux. This is the full build log: https://travis-ci.org/mate-desktop/pluma/jobs/494245847#L5782

However, when you run pluma and start enabling the "Spell Checker" plugin in "Pluma Preferences", you will find that the "Spell Checker" is not enabled and you will see the following warnings, such as https://bugs.archlinux.org /task/61618:
```
(pluma:1139): Gtk-WARNING **: 14:03:42.551: Theme parsing error: <data>:1:17: Expected a string.

(pluma:1139): libpeas-WARNING **: 14:03:51.556: Failed to load module 'spell': /usr/lib/pluma/plugins/libspell.so: undefined symbol: enchant_dict_add_to_pwl

(pluma:1139): libpeas-WARNING **: 14:03:51.556: Error loading plugin 'spell'
```
So 73e0a823b1c040beb68bc48b8279bcef11115a3f fix it.

You are welcome to test and confirm, I hope this bug has been completely fixed.

PR #429, opened 2019-05-20
remove live.gnome from plugins section

selectiveduplicate:
With this I have also removed how we can install third-party plugins... we aren't offering that any more are we?


raveit65:
I think all existing third-party plugins at github aren't ported to gtk3. So no need to offer them.

PR #425, opened 2019-05-12
Python 3

monnerat:
This large PR converts Pluma Python interface and its embedded plugins to Python 3.
It also fixes some Python code bugs and improves removing (although not completely) of Gtk 3 deprecated stuff in Python sources/UIs.

Thanks for considering it.

PR #430, opened 2019-05-20
Reindent all Python sources to ts=4. Strip trailing spaces.

monnerat:
This is the first commit of https://github.com/mate-desktop/pluma/pull/425, as requested in the comments.
It only reindents Python code and strip trailing spaces. No code change thus still working with Python 2.


monnerat:
Thanks for merge

PR #436, opened 2019-05-27
snippets plugin: change code for Python 2 & 3 compatibility

monnerat:
This rewrites the plugin in Python 3, while retaining Python 2 compatibility.

In addition, some bugs are fixed and the Gtk-3 compatibility is improved (mainly regarding UI and widget deprecation).


monnerat:
Thanks for merging :-)

PR #442, opened 2019-06-25
Fix French translation of the "Windows" line ending type

### LINKED ISSUE #441: Weird french translation when saving a file and selecting Windows line ending
#### Expected behaviour

When saving a new file or saving a file as, choices among line endings should be : Unix/Linux | Mac OS Classic | Windows

#### Actual behaviour

When saving a new file or saving a file as, choices among line endings are : Unix/Linux | Mac OS Classic | Fenêtres, which is a literal translation of the plural noun windows.

I personally find this hilarious, however it may obviously be puzzling to whom doesn't already know the feature.

The translation seems to be located in the following file :
https://github.com/mate-desktop/pluma/blob/master/po/fr.po
at line 1574

#### Steps to reproduce the behaviour

Save a new file or save a file as... And seek to select Windows line ending.

#### MATE general version

1.20.4-2

#### Package version

1.20.4-1

#### Linux Distribution

Debian 10 buster

cwendling:
Fixes #441.

This is a very easy fix, and the string is only used in this location so it should have zero side effect.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -1571,7 +1571,7 @@ msgstr "Mac OS Classic"
 
 #: ../pluma/pluma-file-chooser-dialog.c:179
 msgid "Windows"
-msgstr "Fenêtres"
+msgstr "Windows"
 
 #: ../pluma/pluma-file-chooser-dialog.c:442
 #: ../plugins/filebrowser/pluma-file-browser-widget.c:812


raveit65:
Can you please use transifex to fix translations?
https://www.transifex.com/mate/MATE/dashboard/
.....and become a member of you language team.
We cant edit translations here to avoid that a pull from transifex server will override the change.

cwendling:
doesn't transifex sync the other way around as well?

Anyway, it's awfully more work from me, but I tried to join the team there, if it's accepted I'll do it there, but someone else with write access on transifex can do it, it's a very simple change.

raveit65:
No, transifex is like a master for `msgstr` po files.
Mate is only the master for `msgid`
It works like this:
Mate creates .pot file with new `msgid` --> push updated `msgid` to transifex.
Transifex update `msgstr` from updated `msgid` --> Mate pulls updated `msgstr` from transifex.

That means a pull from Transifex overrides every changed `msgstr` in git.

cwendling:
OK I submitted the translation in the master and stable branches over at Transifex, not sure when they'll come up but that should be resolved.

raveit65:
> not sure when they'll come up but that should be resolved.

Next time if someone do a `tx pull -af` which usually happens before a new release.

PR #443, opened 2019-06-25
Fix use of deprecated glib-genmarshal types

cwendling:
This doesn't change any dependency as both non-deprecated replacements `BOOLEAN` and `VOID` were already used in some places.

`NONE` is a deprecated alias for `VOID`, and `BOOL` a deprecated one for `BOOLEAN` (see e.g. the [manpage for glib-genmarshal](https://manpages.ubuntu.com/manpages/cosmic/man1/glib-genmarshal.1.html)).

This gets rid of the warnings like these:
```
WARNING: The token "BOOL" is deprecated; use "BOOLEAN" instead
```


raveit65:
As team member you don't need to fork a repo any more, you can create a new working branch directly ;)

cwendling:
@raveit65 true, but I'm used to the fork-and-pr workflow even on projects where I'm the maintainer :)  I like that being a forked repo not supposed to be upstream for anything but PRs, it has no stability guarantee and I feel rebasing and good force-pushing branches whenever I like.
However if it's more of a problem for you guys for any reason, I can create branches on the master repo no problem.

raveit65:
Of course you can use your repo, it was only a hint. You don't have to use our repo ;)
It's your own decision :)

PR #450, opened 2019-07-10
Fix not running gla11y when not available

sthibaul:


lukefromdc:
Also, someone needs to check this (all of these) with gla11y installed and make sure it works in that case

sthibaul:
CI did test that case

sthibaul:
CI did test that case

lukefromdc:
I don't yet know how to invoke the build options in Meson, which I am still learning to use. Oh yeah it's a lot faster and I don't have to deal with Checkinstall (slow and messy with files it should not pick up), but I'm still learning how to use it.

raveit65:
@lukefromdc 
fixed by https://github.com/mate-desktop/pluma/commit/cdd098013a810ccd8b8baeef3110676f87da1fa9

lukefromdc:
I can confirm a build of 1.23.1 (with https://github.com/mate-desktop/pluma/commit/cdd098013a810ccd8b8baeef3110676f87da1fa9) finished fine

PR #471, opened 2019-07-25
pluma-smart-charset-converter: avoid 'g_type_class_add_private'

sc0w:


lukefromdc:
Does anyone else want to review this and test the character set converter (or know how to test it so I can do it) or should we merge this. We haven't had trouble with this anywhere else but merging code that may not have been actually run is not the best practice

sc0w:
I think this PR can be tested with these steps:

open one file, and open the `"Save As..."` dialog, choose other `"Character Encoding"` and `"Save"`

lukefromdc:
That seems to work fine, though I know little of different character encodings. No crash or surprise on saving to a different character set, and no corruption of the document.

PR #474, opened 2019-07-26
pluma-tab: avoid 'g_type_class_add_private'

sc0w:


lukefromdc:
These have all been troublefree so merging

PR #482, opened 2019-07-31
pluma-file-browser-store: avoid 'g_type_class_add_private'

sc0w:


lukefromdc:
How do I test this for function and invoke the code in pluma-file-browser-store?

sc0w:
@lukefromdc  I think you can test if file browser plugin works as before

PR #490, opened 2019-08-07
pluma-taglist-plugin-panel: avoid 'g_type_class_add_private'

sc0w:


lukefromdc:
OK, I just figured it out: this lives in the side pane, click the + icon to bring it up, double-click one of the offered tags to insert it. Works fine

lukefromdc:
Note that on opening Pluma from a terminal with the taglist sidepane having been showing when it was last closed, we get this warning in the terminal but is also occurs in master and thus has nothing to do with this PR:

`(pluma:30559): Gtk-CRITICAL **: 18:21:25.398: gtk_accel_label_set_accel_closure: assertion 'gtk_accel_group_from_accel_closure (accel_closure) != NULL' failed
`

raveit65:
I saw that warning a while a go in pluma too.

PR #500, opened 2019-08-24
pluma-document: avoid garbarge value in 'file_with_bom' function

sc0w:
Fixes clang analyzer warning:

```
pluma-document.c:682:14: warning: The left operand of '==' is a garbage value
        if ((bom[0] == '\357') &&
             ~~~~~~ ^
```

In the tests, make sure you can't reproduce the issue https://github.com/mate-desktop/pluma/issues/301

### DIFF
--- a/pluma/pluma-document.c
+++ b/pluma/pluma-document.c
@@ -655,6 +655,8 @@ file_with_bom (GFile *file)
 	gchar    bom[3];
 	gchar   *file_path;
 
+	bom[0] = bom[1] = bom[2] = 0;
+
 	file_path = g_file_get_path (file);
 
 	testfile = fopen (file_path, "r");


raveit65:
No idea, if this is the right fix for it, but it seems the warning is gone.

pluma/pluma-document.c
```diff
@@ -655,6 +655,8 @@ file_with_bom (GFile *file)
 	gchar    bom[3];
 	gchar   *file_path;
 
+	bom[0] = bom[1] = bom[2] = 0;
```
cwendling: why not simply initialize the variable above? `gchar    bom[3] = {0};`

cwendling:
A possibly more interesting test is a file with less than 3 bytes, maybe even just for fun:

```console
$ echo -ne '\xEF\xBB' > file.utf8
$ file file.utf8
file.utf8: ISO-8859 text, with no line terminators
$ pluma file.utf8
```

Possibly one out of many times it would think the file has a BOM.

PR #509, opened 2019-10-08
window: Do not override GdkWindowState

vkareh:
When setting the custom pluma window state we override the class method,
but never defer back to the parent class method. This means that window
states like backdrop are never set on Pluma.

### DIFF
--- a/pluma/pluma-window.c
+++ b/pluma/pluma-window.c
@@ -266,7 +266,7 @@ pluma_window_window_state_event (GtkWidget           *widget,
 
 	window->priv->window_state = event->new_window_state;
 
-	return FALSE;
+	return GTK_WIDGET_CLASS (pluma_window_parent_class)->window_state_event (widget, event);
 }
 
 static gboolean


vkareh:
@lukefromdc - how it works is this: the code is overriding the window's `window_state_event` to add its own `new_window_state`. When returning `FALSE`, it tells the window object that that's the end of the code that handles `window_state_event` changes. Instead, this code now returns a call to the _original_ window's `window_state_event` so that GTK can take over and do what it meant to do originally. Think of it as this (pseudo code):
```diff
class GtkWindow {
  function window_state_event() {
    // Do GTK stuff...
  }
}

class PlumaWindow inherits from GtkWindow {
  function window_state_event() {
    // do Pluma stuff...
-    return FALSE; // ignore what GTK intended to do originally
+    return GtkWindow->window_state_event(); // do what GtkWindow intended to do originally
  }
}
```
Not sure if this explains it better or is just more confusing, but that's how I visualize this change. Does that make sense?

lukefromdc:
Oh yeah, thanks for the explanation

PR #512, opened 2019-11-10
Migrate from intltool to gettext

yetist:
This PR is not finished, please don't merge it.

I am having trouble and need some help. I don't know how to write a correct `.its` file for some xml files. These files in `plugins/taglist` direcoty, named `*.tags.xml.in`, for example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<TagList xmlns="http://pluma.sourceforge.net/some-location">
<TagGroup name="XHTML 1.0 - Tags" sort="true">

  <Tag name="Abbreviated form">
    <Begin>&lt;abbr title="</Begin>
    <End>"&gt;&lt;/abbr&gt;</End>
  </Tag>
...
```
I don't know how to make `name` properties in `Tag` translatable, the string is `Abbreviated form`.

@mate-desktop/core-team @mate-desktop/contributors 
Who can help? Or do we need to modify the code to change the xml file structure?

BTW, {it,ru,uk}.po has errors, I temporarily disable them.

PR #513, opened 2019-11-12
add missing format specifier for ngettext

monsta:
This should fix problems with `msgstr[0]` mentioned at https://github.com/mate-desktop/pluma/pull/512. Similar to our other projects, a separate string is used for only one item. One new `msgid` is added.


raveit65:
I guess this can be close now...
If not re-open it.

PR #517, opened 2020-01-08
Fix misuse of full-width symbols

skylee03:
There were some misuse of full-width symbols in the translation of Chinese (China). I fixed it and simultaneously add some new translation in Transifex.


raveit65:
Fixing this at transifex is sufficient. We pull translations from transifex before every new release for ourself.

raveit65:
@skylee03 
You can fix translations at https://www.transifex.com/mate/MATE/dashboard/
Become a member of your language team, please.
A pull from transifex will override everything here at github.

PR #523, opened 2020-02-10
remove note for disabled Python 2 plugins

selectiveduplicate:
Removing the note regarding the disabled Python 2 plugins, since we now have compatibility with Python 3.


raveit65:
We was in freeze and i had already pushed a 1.24 release commit.
Please take a look at git history before you commit and follow discussions.
https://github.com/orgs/mate-desktop/teams/core-team/discussions/20?from_comment=12#discussion-20-comment-12

raveit65:
Np, it's OK :)

PR #526, opened 2020-02-26
Use user page on github when user email is private

rbuj:


raveit65:
Only rebased to fix travis builds.

PR #531, opened 2020-03-31
pluma-commands-help: expand the comment on about dialog

rbuj:
### Before

Screenshot at 2020-03-31 16-58-34

### After

Screenshot at 2020-03-31 17-13-11


rbuj:
The comment is pulled from https://mate-desktop.org/

sc0w:
as I said in mate-terminal, I don't like this change, I prefer a little dialog, for more explanations we have the help

rbuj:
@sc0w following the general directions of gtk https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#GtkAboutDialog--comments

Screenshot at 2020-03-31 22-25-20

PR #542, opened 2020-04-16
pluma-utils: Enable multiline mode on regex search

### LINKED ISSUE #541: Regex library not used in multiline mode
It is possible that this is deliberate, but as a user of text editors and utilities like grep, I would say that..

#### Expected behaviour

^ should match the start of any line, not just the first.
$ should match the end of any line, not just the last.

If I really want the very start of a file, I can use \A for example.

#### Actual behaviour

^ only matches the start of the first line.
$ only matches the end of the last line.

From the documentation, which I now know is at https://developer.gnome.org/glib/stable/glib-regex-syntax.html, this suggests the library is not used in multiline mode, i.e. the G_REGEX_MULTILINE option is unset.

#### Steps to reproduce the behaviour

test.txt
```
11.aaa 222
11 aa 222
 11 aaa 222
11

^\d{2}\.\w{3}+.*
```

regex: `^\d{2}\.\w{3}+.*`

Captura de Pantalla 2020-04-16 a les 18 51 19

_Originally posted by @rbuj in https://github.com/mate-desktop/pluma/issues/538#issuecomment-614772038_

Works! The line is found. But if you move the first line to anywhere else in the file, or insert a different first line before the current one, it doesn't.

#### MATE general version

1.24

#### Package version

1.24

#### Linux Distribution

Ubuntu MATE 20.04 LTS

rbuj:
Test: search `^\d{2}\.\w{3}+.*` in te text below:
```
11 aa 222
11.aaa 222
 11 aaa 222
11.aaa 222
11
11.aaa 222
```

Result: found the even lines

Captura de Pantalla 2020-04-16 a les 22 00 47

Closes #541

### DIFF
--- a/pluma/pluma-utils.c
+++ b/pluma/pluma-utils.c
@@ -1573,7 +1573,7 @@ pluma_gtk_text_iter_regex_search (const GtkTextIter *iter,
 	gboolean found;
 
 	match_string = "";
-	compile_flags = 0;
+	compile_flags = G_REGEX_OPTIMIZE | G_REGEX_MULTILINE;
 
 	if ((flags & GTK_TEXT_SEARCH_CASE_INSENSITIVE) != 0)
 		compile_flags |= G_REGEX_CASELESS;


raveit65:
Looks like we can consider this as Bugfix for cherry-picking to stable branch?

superian:
Thank you.

PR #545, opened 2020-04-20
preferences: enable the plugin 'sort' by default

HacKurx:
Hi,

The "sort" plugin is widely used by users, let's enable it by default.

Thank you.

Best regards.

### DIFF
--- a/data/org.mate.pluma.gschema.xml.in
+++ b/data/org.mate.pluma.gschema.xml.in
@@ -217,7 +217,7 @@
       <description>List of entries in "replace with" textbox.</description>
     </key>
     <key name="active-plugins" type="as">
-      <default>[ 'docinfo', 'modelines', 'filebrowser', 'spell', 'time' ]</default>
+      <default>[ 'docinfo', 'modelines', 'filebrowser', 'spell', 'time', 'sort' ]</default>
       <summary>Active plugins</summary>
       <description>List of active plugins. It contains the "Location" of the active plugins. See the .pluma-plugin file for obtaining the "Location" of a given plugin.</description>
     </key>


raveit65:
Not sure, because it can't be undo when done by accidental.
The only way to undo a sorting is not to save the file, but in this case you can lost other changes.
@mate-desktop/core-team 
Thoughts?

lukefromdc:
I think enabling by default a plugin whose effects cannot be undone is not a good idea unless and until that plugin can be made to support undo.

Yetoo1:
> Since there's a warning on the sort plugin dialogue saying that this can't be undone, I don't think this is going to cause any issue.

Even then, I don't think so since it seems rather incomplete even if it does say it can't be undone. It would be great if a complete feature could exist before pushing it as default to reduce possibility of pushing now, a problem comes up, then a dilemma that comes whether to remove the feature because of the problem or actually fix it and I would hate the decision to remove it (while seemingly unlikely here, I've seen many other projects where smaller problems result in feature removal) because it would be adding unnecessary clicks. It would be great if that path were to be avoided all together by shipping the plugin as default when it is complete. Just my input.

HacKurx:
Hello to all,

Reading the Mate 1.26 changelog for pluma: "The sort plugin now supports undo actions."
To this end, can you please survey your users or study this merge request?

Thanks.

raveit65:
Why should this be a default setting?
I don't think that all users of pluma like to use it.
You can enable it for your self if you need it.

mbkma:
I agree, we should not enable more plugins by default, unless there is a good reason to do so.

PR #558, opened 2020-06-25
Avoid tracking .po files by Git yet managed by Transifex

tamplan:
"git commit -a" can upload .po files with no reason with "git push".
.po files are provides by Transifex synchronization.
I think it's better to avoid this.

### DIFF
--- a/git.mk
+++ b/git.mk
@@ -298,6 +298,7 @@ $(srcdir)/.gitignore: Makefile.am $(top_srcdir)/git.mk $(top_srcdir)/configure.a
 				po/stamp-po \
 				po/.intltool-merge-cache \
 				"po/*.gmo" \
+				"po/*.po" \
 				"po/*.header" \
 				"po/*.mo" \
 				"po/*.sed" \


rbuj:
You can use git command to avoid it (`git checkout -- po/*.po`). Your proposal has other unwanted effects, for that reason I don't agree with your PR. It's just my opinion and other developers may find it useful.

tamplan:
Just to know, could you explain me, please, the unwanted effects you "speak" about?

What is the function of git.mk and .gitignore files? For me is to avoid pull files unwanted to be trackend by git.

Imagine a newbie developer (more or less my case :)) using the command `git status` in case of no use of git.mk and .gitignore files, after a command `./autogen.sh && make` with a lot unwanted files to push.

Before use of git.mk, i use 2 repositories, one every time updated with upstream and and another to test on with use of branches. Before pushing, i use Meld to compare and copy/paste changes to be sure to not push unwanted files.

raveit65:
> I think it's better to avoid this.

And i think it's better to leave it like it is :tomato: 
Except from your personal opinion, why we shoudn't track translation changes at github?
Github is our master and transiflex is not more than a helper.
And please open a report for discussion first if you think we did all wrong before you open a PR, to avoid needless work :P

tamplan:
Ok, understood, discussion open first.

Thanks

raveit65:
> Ok, understood, discussion open first.
> 

Really, I am i doubt that you understood what i said.
Once again, why we shouldn't track translation changes at github?
Github is our master and transiflex is not more than a helper.
So, every code change should be tracked here at github.

tamplan:
Yes i have understood the first time but my sentence is not clear.

Without your explanation, i thank about the translation file changes were tracked on transifex site where there is a historic like here and some PR about translation problems are some time closed indicating to contribute directly on transifex.

I would like to write: I understand your explanation AND the next time i have a doubt about a subject, it's better to open a discussion first by a issue to speak about the subject instead of make a PR perhaps needless (update repository, code review...).

My fault, clear in my minf but not in my message here.

PR #579, opened 2020-10-03
travis: require enchant2

rbuj:


raveit65:
Configure checks for enchant-2 and enchant and i built it always with enchant2-devel for fedora.
But now archlinx build failed.

rbuj:
> But now archlinx build failed.

`make distclean` failed on arch, I guess it's due a file system error:
```
ERROR: files left in build directory after distclean:
./docs/reference/pluma.actions
```

raveit65:
It's weird. the issue started with latest commit in master, but this was Ok in related PR from you.
I've restarted the builds, let see if they failed again.
I rectal that we had this issue in another package a while ago.

raveit65:
Building failed again.
Anyway. maybe there is an issue with autotools in rolling archlinux at the moment.
All this isn't related to PR.
When this issue doesn't goes away we can disable distcheck for archlinux.

PR #580, opened 2020-10-04
Remove pluma-marshal.

mbkma:
Instead use the generic marshaller. Less files, less to maintain.

See:

- https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-new
- http://sourceware.org/libffi/
- https://gitlab.gnome.org/GNOME/gedit/-/commit/373adb7d041b0559953d0ac8cf2213ecc44db956


rbuj (approved):
LGTM

raveit65 (approved):
LGTM

PR #585, opened 2020-11-18
Fix line numbers menu: use g_settings_bind correctly

mbkma:
Test: right-click on line-numbers

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -2247,9 +2247,9 @@ create_line_numbers_menu (GtkWidget *view)
 					gtk_source_view_get_show_line_numbers (GTK_SOURCE_VIEW (view)));
 
 	g_settings_bind (PLUMA_VIEW (view)->priv->editor_settings,
-			 "active",
-			 item,
 			 PLUMA_SETTINGS_DISPLAY_LINE_NUMBERS,
+			 item,
+			 "active",
 			 G_SETTINGS_BIND_SET);
 
 	gtk_menu_shell_append (GTK_MENU_SHELL (menu), item);


lukefromdc (approved):
This fixes a mistake that reversed the order of two variables. With master as of now, if you right click just inside the LEFT EDGE of the view w line numbers enabled, pluma crashes. This can cause loss of work. With this change, that same right-click gets a menuitem to enable or disable line numbers. This works and is a needed bugfix. 

Indents looks the same too, this looks ready to go.

raveit65:
@mbkma 

> Test: right-click on line-numbers

What should happen?
Here behaviour is same as before. No context menu or what ever you expect.
Beside from that pluma builds and runs fine.

mbkma:
@raveit65 before it crashes / after it displays the context menu.

Screenshot at 2020-11-22 13-49-12

rbuj (approved):
Yhe setting's key and the object's property are swapped on g_settings_bind, so there is a segmentation fault.

raveit65:
I get no crash or the context menu w/wo the commit. Maybe HIDPI related?
Any way, fix is fine. I will cherry-pick it to stable.

raveit65 (approved):
LGTM

mbkma:
@raveit65  I don't think this issue exists in 1.24.

raveit65:
Ok, good. 
Cherry-picking doesn't clean apply.

raveit65:
Good news, travis CI builds starting immediately again :) 
https://travis-ci.org/github/mate-desktop/pluma/builds/745209847

rbuj:
culprit 2edb465b646f671f3372c70f020bd3d3076e86ad

PR #587, opened 2020-11-26
Use g_clear_error

rbuj:


mbkma (approved):
Runs and builds fine. From devhelp:
`g_clear_error ():
If err or *err is NULL, does nothing. Otherwise, calls g_error_free() on *err and sets *err to NULL.`
So code should be equivalent.

raveit65 (approved):
Build and runs fine

PR #588, opened 2020-11-28
pluma-message: Fix the position of the inline keyword

rbuj:


mbkma (approved):
builds and runs fine, and position of inline keyword is correct now.

raveit65 (approved):
LGTM

PR #591, opened 2020-12-05
cppcheck warning: The if condition is the same as the previous one

rbuj:


vkareh (approved):
LGTM

This doesn't change the logic, just avoids running the same condition check twice unless necessary.

PR #593, opened 2020-12-10
pluma-utils: Cppcheck warning 'argument order different'

rbuj:


lukefromdc (approved):
This builds and runs fine, code change is very simple A future compiler change could potentially could have turned this warning into a hard failure to build, or into obscure bugs when running the finished binary. 

Warning is gone when running cppcheck, all we get now for this file is some variable scope reduction opportunities, and at line 701 
`pluma-utils.c:701] (warning) Either the condition 'str!=NULL' is redundant or there is possible null pointer dereference: str. [nullPointerRedundantCheck]`
which is for another PR to handle. Including it here would potentially make a future bisect harder

vkareh:
You might want to also re-arrange the comments in `pluma-utils.c`, as the order is wrong there as well:
```c
/**
 * pluma_utils_decode_uri:
 * @uri: the uri to decode
 * @scheme: (allow-none): return value pointer for the uri's
 * scheme (e.g. http, sftp, ...), or %NULL
 * @user: (allow-none): return value pointer for the uri user info, or %NULL
 * @port: (allow-none): return value pointer for the uri port, or %NULL
 * @host: (allow-none): return value pointer for the uri host, or %NULL
 * @path: (allow-none): return value pointer for the uri path, or %NULL
...
```

vkareh (approved):
Other than my comment about documentation, this LGTM

rbuj:
@vkareh done

raveit65 (approved):
LGTM

PR #600, opened 2021-01-03
Update Japanese Translation only

fu7mu4:
Add Japanese Translation only


raveit65:
We are using transifex for our translations.
https://www.transifex.com/mate/MATE/
Please become a member of your language team and update translation there.

PR #607, opened 2021-02-03
Do not build the path

rbuj:


pluma/dialogs/Makefile.am
```diff
@@ -1,12 +1,13 @@
...
-	$(PLUMA_CFLAGS) 					\
-	$(WARN_CFLAGS)
+AM_CPPFLAGS = 					\
+	-I$(top_srcdir)				\
+	-I$(top_builddir)			\
+	-I$(top_srcdir)/pluma 			\
+	-I$(top_builddir)/pluma			\
+	$(PLUMA_CFLAGS) 			\
+	$(WARN_CFLAGS)				\
+	-DPLUMA_DATADIR=\""$(datadir)/pluma"\"
```
mbkma, line 10: Why do you need to define `PLUMA_DATADIR` again? It is already defined in `pluma/Makefile.am`

rbuj: Because `PLUMA_DATADIR` is used in the files below:
```
pluma/dialogs/pluma-encodings-dialog.c: ret = pluma_utils_get_ui_objects (PLUMA_DATADIR "/ui/pluma-encodings-dialog.ui",
pluma/dialogs/pluma-preferences-dialog.c:       ret = pluma_utils_get_ui_objects (PLUMA_DATADIR "/ui/pluma-preferences-dialog.ui",
pluma/dialogs/pluma-search-dialog.c:    ret = pluma_utils_get_ui_objects (PLUMA_DATADIR "/ui/pluma-search-dialog.ui",
```

rbuj: @mbkma Do you want to move `PLUMA_DATADIR` into `config.h`?

mbkma: No, I think it is fine.

Without  `-DPLUMA_DATADIR=\""$(datadir)/pluma"\"`   I now get errors as expected `error: ‘PLUMA_DATADIR’ undeclared`. I did not see those errors at a first try, had to run `make clean && make` to see them.

mbkma (approved):
Pluma builds and runs fine, all items are installed to their proper paths as far as I can see. The `-D` option makes sense here:
```
$ man gcc
<cut>
-D name=definition
    The contents of definition are tokenized and processed as if they appeared during translation phase three in a #define directive.
<cut>
```
Code looks good and removes unneeded complexity.

lukefromdc (approved):
This builds and runs fine. I don't know enough about the build system to evaluate the code changes, but my test build works fine, no surprises or terminal warnings on opening the preferences or search dialogs

lukefromdc:
Calling one more review if someone who knows more about it wants to have a look at it, but I am fine with merging it if not.

raveit65 (approved):
LGTM,
pluma runs fine with that change since several days.

PR #608, opened 2021-02-04
pluma: Use EXIT_SUCCESS macro instead of int value (portability)

rbuj:


yetist (approved):
LGTM

cwendling:
Me being me, I can't help it but mention that `exit(0)` is perfectly equivalent to `exit(EXIT_SUCCESS)`:

[N1570 — Committee Draft — April 12, 2011 — ISO/IEC 9899:201x](http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1570.pdf#[{%22num%22%3A789%2C%22gen%22%3A0}%2C{%22name%22%3A%22XYZ%22}%2C0%2C792%2C0]) (emphasis mine):
> § 7.22.4.4 The exit function
> […] If the value of status is **zero** or `EXIT_SUCCESS`, an implementation-defined form of the status successful termination is returned. […]

raveit65:
> Me being me, I can't help it but mention that exit(0) is perfectly equivalent to exit(EXIT_SUCCESS):

As a system-integrator i don't see why this change is needed :)
Well, it looks like an easy change.
@rbuj @cwendling 
Can you explain why?

rbuj:
For greater portability, you can use the macros EXIT_SUCCESS and EXIT_FAILURE for the conventional status value for success and failure, respectively.
https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

mbkma (approved):
This enhances portability and readability.

cwendling:
> As a system-integrator i don't see why this change is needed :)

It is not needed.

> Well, it looks like an easy change.

It is, the change itself looks perfectly fine.  It's just that it doesn't change *anything* behavior-wise (for any standards-compliant systems, at least).

> This enhances portability

It does not change anything to that regard in a conforming C implementation.  I'm too lazy to check right now, but I'm pretty sure that was already defined as equivalent in C89.

> and readability.

That's subjective and fair.  I have absolutely no problem with plain integers because it's so widely used, but I can see why one would want to use named constants.

---
To summarize: Per C11 (and probably older standards) and POSIX (according to GCC docs), this PR does not change *anything*.  It was fine before, it's fine after.
**I do not have anything against this change**, I just felt the need to mention that it does *not* improve anything, given at least a non-prehistoric environment (if somebody can find a system that doesn't follow either POSIX nor the C standard on that, I'd stand corrected -- and simply yell angrily at that system).

rbuj:
I just want to highlight 2 sentences from the exit(3) man page, which explain why standard constants are preferred to int values.

```
$ man 3 exit
```
> The C standard specifies two constants, EXIT_SUCCESS and EXIT_FAILURE, that may be passed to exit() to indicate successful or unsuccessful termination, respectively.

> The use of EXIT_SUCCESS and EXIT_FAILURE is slightly more portable (to non-UNIX environments) than the use of 0 and some nonzero value like 1 or -1.  In particular, VMS uses a different convention.

PR #609, opened 2021-02-06
fix printing wrap mode

mbkma:
Test:
open `File->Print`, then switch to `Text Editor ` and see that `Text Wrapping` options have an effect now.


rbuj (approved):
LGTM

PR #612, opened 2021-02-06
cppcheck warning: Checking if unsigned expression is less than zero

rbuj:
Test:
```
$ cppcheck --enable=all --quiet . &> cppcheck.log && grep unsignedLessThanZero cppcheck.log 
plugins/taglist/pluma-taglist-plugin-panel.c:419:51: style: Checking if unsigned expression 'strlen(group_name)' is less than zero. [unsignedLessThanZero]
```


mbkma (approved):
This should be equal to `if ((group_name == NULL) || strlen(group_name) == 0)` and dereferencing `group_name` is safe, because left to right evaluation of logical OR is guaranteed.

raveit65 (approved):
LGTM

PR #638, opened 2021-09-05
quickopen-plugin: Use Gio.Settings instead of Pluma.MessageBus.send_s…

fszymanski:
…ync()

This completely avoids the use of the message bus to get the filebrowser
root directory.

Traceback (most recent call last):
  File "/usr/lib64/pluma/plugins/pythonconsole/console.py", line 357, in __run
    r = eval(command, self.namespace, self.namespace)
  File "<string>", line 1, in <module>
AttributeError: 'MessageBus' object has no attribute 'send_sync'

### DIFF
--- a/plugins/quickopen/quickopen/windowhelper.py
+++ b/plugins/quickopen/quickopen/windowhelper.py
@@ -87,22 +87,13 @@ class WindowHelper:
             paths.append(gfile.get_parent())
 
         # File browser root directory
-        bus = self._window.get_message_bus()
+        settings = Gio.Settings.new('org.mate.pluma.plugins.filebrowser.on-load')
+        root = settings.get_string('virtual-root')
+        if root:
+            gfile = Gio.file_new_for_uri(root)
 
-        try:
-            msg = bus.send_sync('/plugins/filebrowser', 'get_root')
-
-            if msg:
-                uri = msg.get_value('uri')
-
-                if uri:
-                    gfile = Gio.file_new_for_uri(uri)
-
-                    if gfile and gfile.is_native():
-                        paths.append(gfile)
-
-        except Exception:
-            pass
+            if gfile and gfile.is_native():
+                paths.append(gfile)
 
         # Recent documents
         paths.append(RecentDocumentsDirectory())


raveit65:
How did you trigger the the traceback ?

fszymanski:
@raveit65 I removed the `try...except` statement from [windowhelper.py:92](https://github.com/mate-desktop/pluma/blob/master/plugins/quickopen/quickopen/windowhelper.py#L92). `Pluma.MessageBus` does not have the `send_sync` method. I tried to use `send_message_sync` just like in [Gedit.py:26](https://gitlab.gnome.org/GNOME/gedit/-/blob/master/gedit/Gedit.py#L26) but it doesn't work.

Example:
```python
bus = self._window.get_message_bus()
# In Pluma `lookup()` returns `Pluma.MessageType` but in Gedit `Gedit.Message`
ts = bus.lookup('/plugins/filebrowser', 'get_root')
# In Pluma this will throw an exception: TypeError: could not get a reference to type class
# Note: In theory, this would return an object of type `Pluma.MessageType`
msg = GObject.new(ts, object_path='/plugins/filebrowser', method='get_root')
bus.send_message_sync(msg)
```
`send_message_sync` requires an object of type `Pluma.Message` and not `Pluma.MessageType`. I would have to do something like this
```python
msg = GObject.new(Pluma.Message, object_path='/plugins/filebrowser', method='get_root')
bus.send_message_sync(msg)
```
but then you get this
```
<string>:1: Warning: g_object_new_is_valid_property: property 'object-path' of object class 'PlumaMessage' is not writable
<string>:1: Warning: g_object_new_is_valid_property: property 'method' of object class 'PlumaMessage' is not writable
```
Note: You can get `Pluma.Message` from `Pluma.MessageType` using the `pluma_message_type_instantiate()` method, but if I'm not mistaken it's not exposed in Python.

If you got any ideas, please, tell me.

mbkma:
Which problem are you trying to solve? Is there a specific bug in current master of the quickopen-plugin?

With your changes the quickopen plugin depends on the filebrowser plugin, which is something I would avoid.

fszymanski:
@mbkma Now this code in `windowhelper.py` is dead.
```python
msg = bus.send_sync('/plugins/filebrowser', 'get_root')

if msg:
    uri = msg.get_value('uri')

    if uri:
        gfile = Gio.file_new_for_uri(uri)

        if gfile and gfile.is_native():
            paths.append(gfile)
```
It always throw an exception (which is  always caught by the `try...except` statement).
The `quickopen` plugin never looks for files in the root directory of the `File Browser Pane`.

PR #651, opened 2021-11-23
Fix double activation of extensions

mbkma:
see https://gitlab.gnome.org/GNOME/gedit/-/commit/aed206a36a561a60fd06db66182f91ad8d666992. In this commit it was deleted as well. The activation of the extensions takes place in `pluma_view_realize` (l. 2316) by calling 
```
		peas_extension_set_foreach (view->priv->extensions,
				            (PeasExtensionSetForeachFunc) extension_added,
				            view);
```
This should fix the problem exposed by https://github.com/mate-desktop/pluma-plugins/pull/31.

### DIFF
--- a/pluma/pluma-view.c
+++ b/pluma/pluma-view.c
@@ -413,11 +413,6 @@ on_notify_buffer_cb (PlumaView  *view,
                       "search_highlight_updated",
                       G_CALLBACK (search_highlight_updated_cb),
                       view);
-
-    /* We only activate the extensions when the right buffer is set,
-     * because most plugins will expect this behaviour, and we won't
-     * change the buffer later anyway. */
-    peas_extension_set_call (view->priv->extensions, "activate", view);
 }
 
 #ifdef GTK_SOURCE_VERSION_3_24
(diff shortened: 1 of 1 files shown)


rbuj (approved):
LGTM. `peas_extension_set_call`  is deprecated and it recommends using `peas_extension_set_foreach` instead.  The extensions are (de)activated without any problems.

```C
/**
 * peas_extension_set_call:
 * @set: A #PeasExtensionSet.
 * @method_name: the name of the method that should be called.
 * @...: arguments for the method.
 *
 * Call a method on all the #PeasExtension instances contained in @set.
 *
 * See peas_extension_call() for more information.
 *
 * Deprecated: 1.2: Use peas_extension_set_foreach() instead.
 *
 * Return value: %TRUE on successful call.
 */
```
https://github.com/GNOME/libpeas/blob/a23df30a290c735a931703d0e87b4f2f9e3a9b3e/libpeas/peas-extension-set.c#L472-L485

raveit65 (approved):
LGTM

PR #653, opened 2021-11-28
pluma-document: Redundant null check

rbuj:


mbkma (approved):
LGTM. `g_free` is NULL safe.

raveit65 (approved):
LGTM

Issue #640, opened 2021-10-04
Revert #!/bin/sh to #!/usr/bin/sh change


jhgit:
In this commit, it was claimed that changing #!/bin/sh to #!/usr/bin/sh was done to "fix shebangs":

https://github.com/mate-desktop/pluma/commit/9a99c3c7b0fbfa9510fb65d96ef70f96f7fd3c62

POSIX does not necessarily specify /bin/sh as THE location for the system's standard shell (I think it allows for it to be in different locations [1]), however it is FAR more common for the standard location to be /bin/sh.

This change breaks FreeBSD and many other systems (probably most all non-linux systems).  [Note: this was noticed now because FreeBSD just updated its pluma port which pulled in this change from 6 months ago]

Can the packager "fix" this to accommodate its location of the standard POSIX shell?  Yes, but I think the fedora script to "mangle" this shebang to use /usr/bin/sh is a poor choice for real world cross-platform compatibility.  I also don't see many (any?) projects heeding this ill-advised (my opinion) change.

It certainly is not POSIX to have /bin and /usr/bin to be the same, so in short, this will break lots of systems that do not follow that practice.

While this change (/bin/sh -> /usr/bin/sh) MAY not merit a "bug" moniker, I feel this part of the change should be reverted as it does not reflect de facto practice.

[1] https://unix.stackexchange.com/questions/416725/does-sh-have-to-be-in-the-bin-directory

_Originally posted by @jhgit in https://github.com/mate-desktop/pluma/issues/626#issuecomment-932966925_

jhgit:
See also this discussion which has links for the POSIX takes on the issue and the mandatory mention of using /usr/bin/env (which has its own set of problems such as location of 'env' and people who might have a different 'sh' in their path, accidentally, maliciously, or otherwise):

https://news.ycombinator.com/item?id=17069408

One take from that discussion is to just not have a shebang line at all which seems to be the most POSIX compliant solution [1] - POSIX seems to be standoff-ish regarding shebangs. I don't know if that would break on any real world systems, but it works on FreeBSD and Linux. Until I see a survey of how real world systems treat the "no shebang" case, I am not recommending that course of action. I still feel /bin/sh is the best choice (certainly better than /usr/bin/sh).

[1] https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

raveit65:
Personal i don't mind because i am able to build pluma for fedora with a patch ;)
[...]

PR #654, opened 2021-12-13
revert /usr/bin/sh shebangs + make search-recursive POSIX compliant

jhgit:
See [issue 640](https://github.com/mate-desktop/pluma/issues/640) for discussion on reverting /usr/bin/sh shebangs.

Also fix search-recursive to be more POSIX compliant.  Remove bashishms and non-POSIX expr(1) usage.  And make the script simpler and more efficient at the same time.


raveit65 (approved):
IDM

Issue #196, opened 2016-08-10
slow searching due to animation


roderickmackenzie:
Hi,
  I've been using pluma every day to write C/python  for the last 2+ years , and up until I've really liked it.  I jumped from gedit to pluma as I could not deal with the new gedit interface.  I've just upgraded from fedora 23 to 24 and started using pluma again.  I've noticed that when I hit crtl+F to do find, the search window pops up just fine.  I type in what I want to find, then hit return.  There is then a noticeable delay (0.2-0.5) seconds between be hitting return and the find dialog closing.  The main text editor window then animates scrolling to the first found item.  The delay in closing the dialog box followed by the time for animated scrolling of the window to the first found item, adds up to a noticeable and pretty irritating delay.  It would be good if the scrolling animation could be turned off and the delay in closing the find dialog removed.  I've temporally switched to kate, but it's not as good as old pluma.  On another note, I also find it difficult to differentiate between tabs as there is no separation between them - although this could be a gtk3/fedora issue.
best,
Rod
##

raveit65:
First,first i switched pluma to gtk3 for f24.
>  I've noticed that when I hit crtl+F to do find, the search window pops up just fine. I type in what I want to find, then hit return. There is then a noticeable delay (0.2-0.5) seconds between be hitting return and the find dialog closing.

Do you mean that window?
![bildschirmfoto zu 2016-08-10 16-52-46](https://cloud.githubusercontent.com/assets/961604/17558468/008b414e-5f1b-11e6-850f-b8f0ce29e7cc.png)
This window does not close if you hit return, it stays always open until you close it with the close button. And i don't see a delay here, but maybe this depends on the document size.
And what do you mean with animated scrolling?

> On another note, I also find it difficult to differentiate between tabs as there is no separation between them - although this could be a gtk3/fedora issue.

This depends on the theme and how notebook tabs are styled.
I guess you mean not the active tab. With most mate themes the inactive tabs get a highlight if you move over with mouse on the tab.
Which theme are you using?

raveit65:
Hmm, you are using Adwaita theme, are you using gnome-shell?
Here in f24 with Mate and a mate theme i can't really reproduce such a delay.
And the tab styling is from adwaita theme.

lsh-0:
+1

this is the first result for "disable find animation in pluma"

[...]

PR #667, opened 2022-11-08
Reduce the enter key delay in the search dialog

robxnano:
This was originally reported in https://github.com/mate-desktop/pluma/issues/196, but there wasn't really a satisfactory outcome. The scrolling animation that was brought up there is only part of the reason for the delay. There's also an animation on the Find button itself which slows down the search process. This makes holding down the enter key to scroll through a large number of results much slower compared to holding down Ctrl+G. This patch bypasses the Find button activation and sends the response signal instantly, which speeds up the process significantly.


lukefromdc (commented):
I didn't notice any animation speed differences with my custom GTK theme
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy
in a back to back test, but this may differ on a per-theme basis. Didn't break the search dialog, but with the "find" button not displaying a reaction (via changing to "pressed") the interface becomes less intuitive. 

Also note that the behavior of other buttons such as "replace" does not change, they become the defaults both with this and with Pluma from master if they were last selected. This lets a user do things like replacing a word one instance at a time w just the return key. They show the highlight and will change appearance to the down/pressed state when return is pressed. Excluding "find" from this makes it seem like the dialog has a responsiveness issue. Removing the animation from ALL the buttons would be still worse, as the user would have no clue what the default is.

In general, any default action in a dialog on striking the return key should show a highlight on that button, and that button should react same as it had been left-clicked when return is pressed. This breaks that pattern

lukefromdc:
Also note that making keyboard-only navigation more difficult by hiding a button's reaction will create real accessability issues. Try using the interface with ONLY the keyboard, as though your mouse just died and you had work to finish or a relative dependent on Orca was using it.

PR #676, opened 2023-05-25
Add quotes before and after font-family's value

kingysu:
1. If the font-family'value is Chinese or any non ASCII characters, the GTK CSS parser will produce a warning message: "Theme parsing error: <data>:1:24: Expected a string.". That is, the GTK CSS parser can not process Chinese or non ASCII characters correctly.
2. So, we should add quotes before and after font-family's value.
3. GTK4 has good support for non ASCII characters.

### DIFF
--- a/pluma/pluma-pango.c
+++ b/pluma/pluma-pango.c
@@ -103,11 +103,7 @@ pluma_pango_font_description_to_css (const PangoFontDescription *desc)
   set = pango_font_description_get_set_fields (desc);
   if (set & PANGO_FONT_MASK_FAMILY)
     {
-      g_string_append (s, "font-family: ");
-      g_string_append (s, "\"");
-      g_string_append (s, pango_font_description_get_family (desc));
-      g_string_append (s, "\"");
-      g_string_append (s, "; ");
+      g_string_append_printf (s, "font-family: \"%s\"; ", pango_font_description_get_family (desc));
     }
   if (set & PANGO_FONT_MASK_STYLE)
     {
--- a/pluma/pluma-pango.c
+++ b/pluma/pluma-pango.c
@@ -104,7 +104,9 @@ pluma_pango_font_description_to_css (const PangoFontDescription *desc)
   if (set & PANGO_FONT_MASK_FAMILY)
     {
       g_string_append (s, "font-family: ");
+      g_string_append (s, "\"");
       g_string_append (s, pango_font_description_get_family (desc));
+      g_string_append (s, "\"");
       g_string_append (s, "; ");
     }
   if (set & PANGO_FONT_MASK_STYLE)


zhuyaliang:
@kingysu It is best to use ```g_string_append_printf```

pluma/pluma-pango.c
```diff
@@ -104,7 +104,9 @@ pluma_pango_font_description_to_css (const PangoFontDescription *desc)
   if (set & PANGO_FONT_MASK_FAMILY)
     {
       g_string_append (s, "font-family: ");
+      g_string_append (s, "\"");
       g_string_append (s, pango_font_description_get_family (desc));
+      g_string_append (s, "\"");
       g_string_append (s, "; ");
```
cwendling, line 110: As @zhuyaliang suggests, and it could actually improve the whole thing:
```suggestion
      g_string_append_printf (s, "font-family: \"%s\"; ",
                              pango_font_description_get_family (desc));
```

kingysu:
> @kingysu It is best to use `g_string_append_printf`

Got it.

zhuyaliang (approved):
Two warnings confirmed that one has been fixed, so we agree to merge

kingysu:
> Two warnings confirmed that one has been fixed, so we agree to merge

What are the two warnings?

zhuyaliang:
```Edit->Preferences```

```(pluma:911570): Gtk-WARNING **: 13:46:50.973: Theme parsing error: <data>:1:17: Expected a string.```

File: pluma/pluma-pango.c [hunk 2]
```diff
@@ -103,9 +103,14 @@ pluma_pango_font_description_to_css (const PangoFontDescription *desc)
   set = pango_font_description_get_set_fields (desc);
   if (set & PANGO_FONT_MASK_FAMILY)
     {
-      g_string_append (s, "font-family: ");
-      g_string_append (s, pango_font_description_get_family (desc));
-      g_string_append (s, "; ");
+      const char* font_familay = pango_font_description_get_family (desc);
+      if (font_familay[0] != '"') {
```
zhuyaliang, line 107: Why add the ``` if (font_familay[0] != '"')```

kingysu: Because `font-family: "Sans"` is valid. So if we does  not check if there is a quote, we will make new CSS code like this: `font_familay: ""Sans""`

kingysu:
> `Edit->Preferences`
> 
> `(pluma:911570): Gtk-WARNING **: 13:46:50.973: Theme parsing error: <data>:1:17: Expected a string.`

Yes, this is the warning that I methioned before.

pluma/pluma-pango.c
zhuyaliang: How can I generate font ```family: "Sans"```.I cannot set the font to ```"Sans"```

kingysu: Sorry, I have made a mistake. I have tested that the font family name from Pango is without quotes.
So, I have reverted the code.

kingysu: 1. In pluma, the font name is without quote by default.

2.  We could use the font name without quote to call pango_font_description_from_string function in order to get a PangoFontDescription object.  Therefore, if we use the previous `PangoFontDescription `object to call `pango_font_description_get_family `function, we will get a font family name without quote.

3. We could use a font name with quote to call pango_font_description_from_string function in order to get a PangoFontDescription object.  Therefore, if we use the previous `PangoFontDescription `object to call `pango_font_description_get_family `function, we will get a font family name with quote.

4.  So, we could modify `pluma_pango_font_description_to_css `function in order to have less effect on other codes.

zhuyaliang (approved):

kingysu:
I have comitted two patches to GTK3 to fix the problem and the two patches have been merged. 
PR1: [Make sure that the charcode is signless](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6030); 
PR2: [Fix a typo](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6040)

PR #685, opened 2023-09-21
Captions/Sub-Captions translated.

bastianovic-dev:


raveit65:
We use transifex for our translations. https://app.transifex.com/mate/MATE/dashboard/
Please translate it there.
Thank you.

PR #690, opened 2023-11-27
modelines plugin: add a shorter version of the 'textwidth' option ('tw')

fszymanski:


raveit65:
Please rebase with master to fix CI building.

raveit65:
I do not use the modul for myself. How can this be tested?

fszymanski:
`modelines` allow you to set variables specific to a file. If you put the following in the last line of a Python program, you would get a textwidth of 60 chars when editing that file:
```
# vim: tw=60
```
In the case of `Pluma` the `Display right margin at column` option will be set to 60.
See: https://www.cs.swarthmore.edu/oldhelp/vim/modelines.html

raveit65 (approved):
LGTM,
both versions `# vim: tw=60` and `# vim: textwidth=60` are working now.

PR #708, opened 2025-01-30
Improve window behaviour

mdeslaur:
These two commits improve the window size behaviour.

The window currently grows bigger and bigger, as it is using the window allocation as the saved size, which isn't recommended per the gtk documentation.

In addition, the maximized window saving behaviour is improved by saving the unmaximized size, which is a more consistent user experience.

### DIFF
--- a/pluma/pluma-app.c
+++ b/pluma/pluma-app.c
@@ -552,23 +552,18 @@ pluma_app_create_window_real (PlumaApp    *app,
         GdkWindowState state;
         gint w, h;
 
+        g_settings_get (app->priv->window_settings,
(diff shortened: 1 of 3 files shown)


lukefromdc (approved):
This fixes a nasty problem with Pluma under wayland. Never noticed this behavior under Xorg but I get it big-time in the wayland session with wayfire and the server using CSD to draw SSD style decoration. Now pluma reopens at the same size I last closed it instead of growing rapidly from the side of CSD decoratipns even when identical to my Marco theme

lukefromdc:
Simpler, cleaner coding too.

PR #718, opened 2025-10-16
pluma-window: Add right-side panel support

vkareh:
Add a third panel on the right side of the editor window, enabling a three-panel layout with left, right, and bottom panes.

The right panel is hidden by default and its state persists across sessions. The main use case is enabling plugins for previews, document outline viewers, or tag lists, similar to other IDEs.

As an initial example, I've moved the `taglist` plugin to the right pane. Other examples to be available in https://github.com/mate-desktop/pluma-plugins/pull/41


mbkma:
Very nice, cool feature! I tested it and it works fine :) Only one thing: The right-side panel size does not seem to get restored on restart. The size of the left-side panel does.

PR #738, opened 2026-04-02
externaltools: fix python invalid escape sequence warning

vkareh:
In later versions of Python, invalid escape sequences changed from a DeprecationWarning to SyntaxWarning. This change uses a raw string for the RE_KEY regex pattern to fix the invalid escape sequence warning.

Fixes #737

### DIFF
--- a/plugins/externaltools/tools/library.py
+++ b/plugins/externaltools/tools/library.py
@@ -197,7 +197,7 @@ class ToolDirectory(object):
 
 
 class Tool(object):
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
While I don't know much python and can't evaluate the escape sequence change as code, installing by dropping the file into my existing pluma install didn't break the "remove trailing spaces" external tools plugin, which worked.

Since the python change is a reported breaker, we do need to deal with this

mbkma (approved):

